PR #130, opened 2014-09-12
Allow editable option also in 'form' mode

PekkaPLeppanen:
Currently in `tree` mode the structure can be changed, but editable option works and in `form` mode the structure is read-only, but editable option is omitted.

I needed a mode, where the structure is read-only and I could also programmatically decide is the property or value editable.


josdejong:
This should be fixed now in v5.0.0

josdejong:
Sorry it wasn't fixed yet, I've fixed it in the develop branch now

PR #153, opened 2015-01-09
In case of a click listening div under the jsonEditor component

maalej:
I have a div listening to click events behind the jsonEditor. So when I click on the jsonEditor component, both of the divs catch that event.
So:
event.stopPropagation(); 
Solved my problem

### DIFF
--- a/jsoneditor.js
+++ b/jsoneditor.js
@@ -806,6 +806,7 @@ return /******/ (function(modules) { // webpackBootstrap
 	      if (target.nodeName == 'BUTTON') {
 	        event.preventDefault();
 	      }
+	      event.stopPropagation();
 	    };
 	    this.frame.oninput = onEvent;
 	    this.frame.onchange = onEvent;


josdejong:
Thanks for your idea. 

What is exactly your use case where you have a click handler in some element behind the JSON Editor?

maalej:
I am creating a component which is a mixture of this example: http://mbostock.github.io/d3/talk/20111018/partition.html and the jsonEditor and I need to superpose the jsonEditor in front of the visualization. 
So for example when I click on the jsonEditor the visualization behind catches also my click. but with my commit it doesn't.

It would be gratefull if you merge it.
Thanks

josdejong:
Ok thanks for the clarification. The thing is, I'm not sure whether it's a good thing to stop propagation of click events by default. There may be cases as well where you actually *do* want to catch this click event outside of the editor. By stopping propagation, only one of these use cases can be served, while, when not blocking the event, both cases can be realized.

There is an easy solution for your case where you do *not* want to have the click propagated: add a click handler to the HTML container where you created the JSON Editor, which stops further propagation. Something like:

```js
var container = document.getElementById("jsoneditor");
container.onclick = function (event) {
  event.stopPropagation();
};

var editor = new JSONEditor(container);
``` 

Does that make sense?

maalej:
yes thank you

josdejong:
Ok thanks

PR #159, opened 2015-01-27
Handle changed node types

fisadev:
Don't crash when receiving a value with different structure than the type info. Try to recover as much as we can,  and fill the needed incompatible values with new default values. Also inform of the errors using editor.option.error()


josdejong:
Hi Juan, thanks for your PR. However this PR looks like it will override JSON Editor completely with AIMARA Editor, rather than doing some small fixes/improvements which I think was your intention?

fisadev:
Ouch! Sorry! This was meant to be a pull request to the aimara editor master branch, not to your json editor. My bad!

josdejong:
Ah ok, that expains it :)

Issue #144, opened 2014-11-05
feature request: Strip to {} and Strip to []


mvark:
JSON Viewer - http://jsonviewer.codeplex.com/ is a desktop tool that has a feature "Strip to {}" and "Strip to []". This option removes the function name, comments and brackets so that there is no JSON validation error. 

For example, this JSON feed will throw a parsing error with your viewer - https://ajax.googleapis.com/ajax/services/search/news?q=Hyderabad,India&v=1.0&rsz=5&callback=jQuery18205345600936561823_1415180407646&_=1415180677951

I have not seen the feature of stripping JSON to exclude anything outside the main opening and closing braces, in any of the online JSON viewers. Can you please add this feature?

JosDet:
Fuck！！Big Fat liar! Hurry to repay $500000!!!!!

PR #162, opened 2015-02-10
Feature #144

yanivefraim:
This is a fix for feature #144. It will remove JSON's surrounding characters. For example:
`XYZ{"a":"b"}ZYX => {"a":"b"}`

This is useful for "stripping" JSONP responses to be a valid JSON:
`/* callback */jQuery18205345600936561823_1415180407646({"responseData": {"results":[{"GsearchResultClass....}) `
will become: 
`{"responseData": {"results":[{"GsearchResultClass....}`


josdejong:
Thanks for your pull request Yaniv. Few remarks:

- Can you do the pull request the `develop` branch instead of `master` please? (I should add a file `CONTRIBUTING.md` to the project, will do that)
- Your current solution of removing all characters befor the first `{` and after the last `}` will work in most cases (when JSON is an object), but not if the JSON is an array, string, number, boolean, or null (which is all valid JSON). 
- Maybe it's easier to split this task in two:
  - first remove all comments like `/* ... */` from the code
  - then, remove JSONP notation (i.e. the code starts with a function name, then a bracket open, and ends with a bracket close, all possibly having whitespaces before/after). That should work with all JSONP code, not just of jQuery.

yanivefraim:
Sure, I'm on it

PR #164, opened 2015-02-11
Feature #144

yanivefraim:
This is a fix for feature #144. It will remove JSONP surrounding:
`/* callback */ jQuery18205345600936561823_1415180407646 ( {"responseData": {"results":[{"GsearchResultClass":"abc"}]}} ); `
will become: 
`{"responseData": {"results":[{"GsearchResultClass":"abc"}]}}`

The steps for stripping the JSONP object: 
1. Remove all comments like /* ... */ from the code
2. Remove JSONP notation (i.e. the code starts with a function name, then a bracket open, and ends with a bracket close, all possibly having whitespaces before/after). That should work with all JSONP code, not just of jQuery.


josdejong:
Thanks, looks good. 

Maybe the regexp for JSONP can be refined a little bit. For example it allows spaces in function names, and allows garbage before the function name and after the function end bracket.

Not fully tested, but something like this:

```
^\s*[\dA-z_$]+\s*\(([\s\S]*)\)\s*;?\s*$

^                  start of document
\s*                zero or more whitespaces
[\dA-z_$]+         a function name
\s*                zero or more whitespaces
\(                 bracket open

([\s\S]*)          one or more of any character including newlines
                     -> this is the part we want to extract

\)                 bracket close
\s*                zero or more whitespaces
;?                 optionally a semicolon
\s*                zero or more whitespaces
$                  end of document
```

With this regexp you immediately get the inner JSON of the JSONP function as second argument of the match:
```js
// test whether JSON is enclosed by a JSONP function
var match = jsonString.match(/^\s*[\dA-z_$]+\s*\(([\s\S]*)\)\s*;?\s*$/);
if (match) {
  var jsonString = match[1];
}
```

yanivefraim:
Awesome - this version is more elegant and works great. (did some testing and verified that it works)

josdejong:
Ha ha nice teamwork this way ;)

yanivefraim:
Awesome, thanks!

PR #170, opened 2015-03-18
Support for custom themes.

nfvs:
This PR adds support for specifying custom themes in JSONEditor options when using the code editor.

Since only the 'jsoneditor' theme is included by default, using this requires you to include the theme yourself _outside_ of jsoneditor. Here's an example using browserify:

```javascript
require('brace');  // required for the next line to work
require('brace/theme/monokai');  // load the theme

var JSONEditor = require('jsoneditor');
var editor = new JSONEditor(el, { mode: 'code', theme: 'ace/theme/monokai' });
```


josdejong:
Thanks Nuno, a nice and simple solution.

PR #180, opened 2015-04-14
Put value type colors in css instead of hardcoding

artch:
This was described in `TODO` comment [here](https://github.com/josdejong/jsoneditor/blob/master/src/js/Node.js#L1027). Now it is possible to theme these colors using only CSS, without modifying the JS code.


josdejong:
Thanks for the PR @artch !

A few things:

- Can you please do the pull request to the `develop` branch?
- No need to commit the files in `/dist`, they will be auto generated when a new version is released. (see [CONTRIBUTING.md](https://github.com/josdejong/jsoneditor/blob/master/CONTRIBUTING.md)).
-  Right now you use `util.removeClassName` (5x) and `util.addClassName` (1x) for every value. That's quite some DOM interaction, which is expensive. How about a simple, single `domValue.className = newClassName`?

1j01:
:+1: To restyle this for a dark theme I'm having to do some hacks:
```css
.jsoneditor .value[style*="color: green;"] {
	color: lime !important;
}
.jsoneditor .value[style*="color: red;"] {
	color: rgb(75, 193, 255) !important;
}
```

josdejong:
Thanks @1j01 . I should pick this up again.

josdejong:
I've implemented this now in the develop branch

PR #183, opened 2015-04-27
Allowing arrow keys to move across objects

MarioMey:
Maybe, this could be an option. Because, actually, arrow keys are used to move across the strings. I would like to move across the objects to reach the one I want to edit and there, I press Enter, for example, to edit it.

PS: I don't understand well this "Pull request" and the comparing branches. Sorry if I did it wrong.


MarioMey:
I'm seeing that this is in the wrong WRONG place. I'm very sorry.

I read "pull request" and, translated to my language, it is like "suggestion". I didn't find "suggestions" section! (maybe I should send you a mail...)

MarioMey:
Also, I don't know how to remove this pull request...

MarioMey:
ALSO, I just discover about Alt-arrow keys... I feel very ashamed! I'm very sorry.

josdejong:
Wore welcome. So I guess your pull request is redundant?

MarioMey:
Yes, absolutly.

I'll use this software with my project... maybe I'll have more 
suggestions or bugs-reports. Where should I report them?

For example:
*
Suggestion:* I really like the arrange of http://jsoneditoronline.org/, 
but it is not in examples (with both views at the same time and 
automatic windows sizes). Maybe it could be added...?

*Bug:* when editting in "Tree" mode, the file is saved no-beatifulled. 
To make beautifull files, I have to change to "Code" mode and save from 
there. Could the files be saved always beatuifulled?

If it would be necessary, I can report these suggestion and bug in 
corresponding reporting site...

Thank you very much.



El 27/04/15 a las 09:28, Jos de Jong escibió:

josdejong:
You can post suggestions and bug reports here in the issues section of the project (https://github.com/josdejong/jsoneditor/issues).

I'm not planning on adding a two panel example but you can of course create it yourself if you like.

There are two methods to retrieve the JSON from the editor:
- `getText()` returns a (non-formatted) string with the JSON contents.
- `get()` returns the JSON object itself, which you can stringify yourself. If you want to do that formatted, you can do so like:
  ```js
  var str = JSON.stringify(myEditor.get(), null, 2);
  ```
  or something like that.

PR #243, opened 2015-12-12
Use SVG for icons

1j01:
The context menu icon is now transparent.
The plus icon is pixel-aligned.
On the fifth and sixth rows, the arrow paths are merged.
Some other things that were not defined with integers now are.
The magnifying glass is arbitrarily changed on account of me redoing everything. It's a bit smaller now.
The red and green colors are changed a bit.


1j01:
Oh, don't forget to delete the `png` from `dist` in the next release because the build won't clean that up.

josdejong:
Thanks a lot Isaiah, nice work! Next week I have time to do some other improvements and after that I will do a release.

Issue #255, opened 2016-01-15
Missing 'search' option in v5.1.0


MiroHibler:
I'm getting console warning about 'search' option being unknown.

josdejong:
Thanks for reporting this. The option still works but is wrongly displaying a warning in the console. The new options validator was missing the option `search`. Fixed now in the develop branch.

MiroHibler:
Yeah, I was about to submit a PR, but got stuck with my GitHub SSH permissions for some yet unknown reasons... :)

josdejong:
thanks anyway!

PR #256, opened 2016-01-15
Error if 'search' option set to false

MiroHibler:
Fixed error "Cannot read property 'clear' of undefined" thrown if `search` option set to `false`.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -3,6 +3,11 @@
 https://github.com/josdejong/jsoneditor
 
 
+## not yet released, version 5.1.1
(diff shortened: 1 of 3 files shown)


josdejong:
Thanks for the fix! 

I see I accidentally pushed my fix of #255 to master instead of develop ::shame::. And I should have actually tested the `search` option after fixing the warning...

MiroHibler:
No sweat, we're all been there ;)

josdejong:
Friday evenings are dangerous...

PR #277, opened 2016-02-29
i18n for validation error messages and add lang option

bearmini:
Add i18n for validation error messges using ajv-i18n.
Also add 'lang' option to choose which language will be used for the i18n.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
   },
   "dependencies": {
     "ajv": "3.4.0",
+    "ajv-i18n": "1.0.0",
     "brace": "0.7.0"
   },
   "devDependencies": {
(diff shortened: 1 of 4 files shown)


josdejong:
Thanks Takashi, I will review your PR as soon as possible.

josdejong:
Your implementation looks neat,thanks. 

I was thinking however, before introducin a new option like `lang`, we should have full support for localization in the JSON Editor (context menu, tooltips, JSON errors, etc), not just for the validation errors. Also, like the `ajv-i18n` is separate from `ajv` itself, I think it will be best to create a `jsoneditor-i18n` library separate from `jsoneditor` itself. What do you think?

Issue #75, opened 2013-10-31
Custom items in contextual menu


tdalous:
I would like to know which way would be the best to customize the contextual menu, to add specific items for instance.. You can find the idea illustrated attached, thanks.
![capture](https://f.cloud.github.com/assets/4903839/1445020/06d5d2fe-4217-11e3-9717-3fe60a8418b3.jpg)

PR #284, opened 2016-03-24
Dynamic Context Menu

tdakanalis:
Define the nodes that will have a context menu and which actions are permitted for each node by defining the 'context' option in the editor's options. 

For example the options below define that only node1, node2, and all the children of node3 and node4 will have a context menu. Moreover, node1 and node2 allow only the 'Remove' action while the children of node3 and node4 allow both the 'Duplicate' and 'Remove' actions:

```
context: {
    items: { 
        node1: ['Remove'], 
        node2: ['Remove'] 
    }, 
    children: { 
        node3: ['Duplicate', 'Remove'], 
        node4: ['Duplicate', 'Remove'] 
    } 
}
```

### DIFF
--- a/src/js/JSONEditor.js
+++ b/src/js/JSONEditor.js
@@ -75,7 +75,7 @@ function JSONEditor (container, options, json) {
     if (options) {
       var VALID_OPTIONS = [
         'ace', 'theme',
-        'ajv', 'schema',
+        'ajv', 'schema', 'context',
         'onChange', 'onEditable', 'onError', 'onModeChange',
         'escapeUnicode', 'history', 'search', 'mode', 'modes', 'name', 'indentation'
       ];
--- a/src/js/Node.js
+++ b/src/js/Node.js
@@ -1433,13 +1433,38 @@ Node.prototype.getDom = function() {
     }
     dom.tr.appendChild(tdDrag);
 
-    // create context menu
+    //Get the context menu options
+    var contextOptions = this.editor.options ? this.editor.options.context : undefined;
+    //Get the items with a defined context menu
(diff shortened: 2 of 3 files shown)


josdejong:
Thanks Themistoklis. This looks interesting. 

Some thoughts:

- It may be a good idea to look a bit broader. There is a long standing feature request to allow for custom entries in the context menu, see #75. Maybe we can pick this up together with restricting the allowed actions in the context menu.

- There are shortcuts to create/duplicate/remove entries. I guess we should disable these too to truly restrict options.

- The current configuration where you can enter "items" and "children" may be a bit restrictive. Does "children" hold only for direct children or also recursively for nested children (grandchildren)?
  It can be interesting to look into how JSONEditor currently allows to configure editability of fields and values, and use a similar solution to allow/restrict actions on nodes by providing a callback function:

  > {function} onEditable
  > Set a callback function to determine whether individual nodes are editable or read-only. Only applicable when option mode is tree. The callback is invoked as editable(node), where node is an object {field: string, value: string, path: string[]}. The function must either return a boolean value to set both the nodes field and value editable or read-only, or return an object {field: boolean, value: boolean}.

  Maybe we could reuse the existing `onEditable` and extend it with options to return "actionable" properties as well as "editable" properties. Or maybe we should make a new option `onActionable` or something like that.

tdakanalis:
Yes, definitely a completely dynamic context menu would be much more useful! We can proceed the way you suggest! I prefer we to introduce a new option, instead of using the onEditable, in order to keep it clear.

josdejong:
Agree

dhughes-xumak:
#438 adds support for custom context menu items.

Issue #282, opened 2016-03-23
Feature request: add options menu to items with enum in schema


maaikez:
I have a feature request: I would like to have an options menu for items with an enum in the schema. 
So if this is (part of) the schema:
	"color": {
		"type": "string",
		"minLength": 1,
		"enum": [
			"black", "red", "green", "orange"
		]
	},

It would be really nice to have an options menu for this item (like you have the checkbox for booleans).

I was wondering how the fields and values are linked with the schema (tried to do this myself), but I didn't find out yet. Or do I have to iterate over the items in the schema for every value?

josdejong:
Yes that idea crossed my mind too. Thanks for your suggestion.

PR #285, opened 2016-03-24
Display Enums As Drop Downs

tdakanalis:
The tree editor displays now the enum types with a drop down. The only you have to do is to define the enum values in the json schema.  (see feature request #282). 

Example:

```
"properties": {
    "fruits": {
        "enum": ['orange', 'apple']
    }
}
```


josdejong:
This looks promising, Thanks!

A few remarks:

- I couldn't get the select box working so far. I would expect example [07_json_schema_validation.html](https://github.com/josdejong/jsoneditor/blob/master/examples/07_json_schema_validation.html) to work right away and show a select box for the field gender but it doesn't ? Or am I overlooking something obvious?
- Does this also work with enumerations in nested properties?
- As far as I know `enumLabels` is no part of JSON Schema, is that correct? I doubt whether it's a good idea to use non-standard stuff in JSON Schema. I think just having the values themselves is enough. And if we really want we could implement a way to pass these labels via the configuration of JSONEditor instead of having them in the JSON Schema.
- I'm not sure how we should handle invalid values. Allow them or not? Select box doesn't allow invalid values and will probably adjust invalid input without the user wanting too. JSON Schema has a *lot* of ways define valid input (like input must be a positive number), and so far JSONEditor does not restrict what you can enter, only give you a warning when it's invalid.

What do you think?

tdakanalis:
Here are my answers:
- I have applied some fixes and now you should be able to see the drop down.
- Yes this works for enumerations in nested  properties
- Yes you are right, enumLabels is not part of the json schema. I just added that extra ability because a lot of people would be happy to have it. Do you have any suggestions about how we could define these labels in JSONEditor's options?
- The selected value of the drop down is still being validated on change event and the warning will be still displayed.

josdejong:
Ah, now I can try it out. This is really cool :)

I came across an issue that the selectbox isn't created or removed when the field is changed. Could you have a look into that? To repeat: Open the example 07_json_schema_validation.html, change the field "gender" to "foo". Now the select box should be removed but isn't. After renaming the field to foo, append a new node, and change the field to "gender". Now the value should become a select box but this isn't the case.

As for the labels: I know this will be more complicated to implement but I really think it's worth keeping a strict separation. Maybe we could create a new option `enums` where you can enter objects with enums. The keys can be a JSON path to the enum, and the value can be an object where the keys are the names of the enums as they are represented in the JSON schema, and the values can be the label to be displayed in the select box:

```js
var options = {
  enums: {
    '.gender': {
      male: 'Male',
      female: 'Female'
    }
  }
}
```

and when the enum is nested somewhere:

```js
var options = {
  enums: {
    'clients[*].gender': {
      male: 'Male',
      female: 'Female'
    }
  }
}
```

There is a [function `util.parsePath` in src/util.js](https://github.com/josdejong/jsoneditor/blob/develop/src/js/util.js#L633-L670) which should make it easy to parse the paths and match it against the path of the current Node, `node.getPath()`. I've just extended util.parsePath to support the wildcard character `*` too, and did a fix in Node.getPath (formerly named Node.getFieldsPath) in the develop branch so you will have to merge the latest two commits first.

tdakanalis:
I have fixed the issues you mentioned. Regarding the labels, I will make the changes you suggest and I will let you know when it is ready.

josdejong:
Thanks! 

Can you for now please remove or comment the `enumLabels` related code?

tdakanalis:
Good Morning!

I have removed the enumLabels related code.

josdejong:
Thanks Themistoklis!

josdejong:
This is really nice :)

maaikez:
Looks good! However, I use 'type' on different locations, not all with the same possible enum values, but I get the same enum values for every type.

For example: I have a house with property type and enum values appartment, bungalow, villa, terraced house. And I have a roof type with enum values flat, gabled. I get the same values for both type dropdowns, so one of them is not giving the right values.

I also have the validation on, and when choosing one of the wrong types, I of course also get a validation error :).

josdejong:
@maaikez thanks for your feedback. Could you maybe post a minimal example of a JSON document and schema where this issue occurs?

maaikez:
Yes, here you are:

`{
	"airplanes": [{
		"name": "cessna",
		"motor": {
			"type": "jet a1",
			"cooling": "air"
		},
		"wings": {
			"type": "braced"
		},
		"type": "high wing"
	}, {
		"name": "superdimona",
		"type": "low wing",
		"motor": {
			"type": "avgas",
			"cooling": "water"
		},
		"wings": {
			"type": "monocoque"
		}
	}]
}`


Schema:

`{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "description": "",
  "type": "object",
  "properties": {
    "airplanes": {
      "type": "array",
      "uniqueItems": true,
      "minItems": 1,
      "items": {
        "required": [
          "name",
          "type"
        ],
        "properties": {
          "name": {
            "type": "string",
            "minLength": 1
          },
          "motor": {
            "type": "object",
            "properties": {
              "type": {
                "type": "string",
                "minLength": 1,
                "enum": [ "jet a1", "avgas" ]
              },
              "cooling": {
                "type": "string",
                "minLength": 1,
                "enum": ["water", "air"]
              }
            },
            "required": [
              "type",
              "cooling"
            ]
          },
          "wings": {
            "type": "object",
            "properties": {
              "type": {
                "type": "string",
                "minLength": 1,                
                "enum": [ "monocoque", "braced" ]
              }
            },
            "required": [
              "type"
            ]
          },
          "type": {
            "type": "string",
            "minLength": 1,
            "enum": [ "high wing", "low wing"]
          }
        }
      }
    }
  },
  "required": [
    "airplanes"
  ]
}`


All 'type' dropdowns are jet a1/avgas.

josdejong:
Thanks, going to check it out.

maaikez:
Thank you! 
I'm sorry for asking, but when do you think you can look at this? Because it's quite annoying (json editor is kind of unusable for the enums now) and if it akes too long I have to find another way.

josdejong:
I hope we will have this fixed within a few weeks.

@tdakanalis do you have time to look into this issue with the new enums dropdown?

tdakanalis:
Yes, sure. I will fix it until the end of the week.

josdejong:
Great, thanks!

maaikez:
Thanks a lot!

josdejong:
@tdakanalis any news on this?

tdakanalis:
I have created the pull request #300 that contains a fix on this issue. However, the json schema attribute "id" must be defined for each type node that will be in the following form:

{
    "$schema": "http://json-schema.org/draft-04/schema#",
    "description": "",
    "type": "object",
    "properties": {
      "airplanes": {
        "type": "array",
        "uniqueItems": true,
        "minItems": 1,
        "items": {
          "required": [
            "name",
            "type"
          ],
          "properties": {
            "name": {
              "type": "string",
              "minLength": 1
            },
            "motor": {
              "type": "object",
              "properties": {
                "type": {
                  "id": "airplanes/motor/type",
                  "type": "string",
                  "minLength": 1,
                  "enum": [ "jet a1", "avgas" ]
                },
                "cooling": {
                  "type": "string",
                  "minLength": 1,
                  "enum": ["water", "air"]
                }
              },
              "required": [
                "type",
                "cooling"
              ]
            },
            "wings": {
              "type": "object",
              "properties": {
                "type": {
                  "id": "airplanes/wings/type",
                  "type": "string",
                  "minLength": 1,
                  "enum": [ "monocoque", "braced" ]
                }
              },
              "required": [
                "type"
              ]
            },
            "type": {
              "id": "airplanes/type",
              "type": "string",
              "minLength": 1,
              "enum": [ "high wing", "low wing"]
            }
          }
        }
      }
    },
    "required": [
      "airplanes"
    ]
  }

maaikez:
Thank you for the fix!

However, is id not meant for something else?
http://json-schema.org/latest/json-schema-core.html#anchor27
"The "id" keyword (or "id", for short) is used to alter the resolution scope. When an id is encountered, an implementation MUST resolve this id against the most immediate parent scope. The resolved URI will be the new resolution scope for this subschema and all its children, until another id is encountered."

If checking the schema against the json is working, why is it not possible to use the same schema to check the enum? How is the schema validation done? 
Isn't it possible to implement this without updating the schema (in a way it is still possible to use 'id' the way it is supposed to be used)?

josdejong:
hm, `id` should indeed not be necessary I think.

josdejong:
@maaikez I've fixed this in v5.5.4. Sorry for all the delays.

@tdakanalis it would be great if you could have a look at my fix and see whether I didn't overlook something. See the following commit: https://github.com/josdejong/jsoneditor/commit/b922aceb07c4a714aaa310af6a281f35b7b9db8a.

maaikez:
Thank you for fixing it! I'm going to take a look and try it out!

maaikez:
Thank you! A quick look shows it is working better.

However, I found two more issues (sorry): 
- When the editor has a name as option, the first item in the path array is the name of the editor. In the Node._findSchema function, in the for loop, when i = 0, var key = path[i]; is the name of the editor. Then 
childSchema = childSchema.properties[key] || null
childSchema is always null. So when you give the editor a name, the enum dropdowns are unfortunately not working.

I fixed this by checking if the editor has a name, if so, I start the for loop from 1, otherwise from 0.

- When using 'allOf', the enums (deeper) in the 'allOf' are not found (the code is not looking in the allOf definitions?).

josdejong:
Thanks for your feedback! I've fixed your first issue in the develop branch (see https://github.com/josdejong/jsoneditor/commit/b67d683f1a73fa0f38018d418331ab22ca41127a).

I'm not sure how to reproduce your second issue. Having an object `anyOf`, `allOf`, or `oneOf` containing an `enum` field should work. Things work fine for example when I replace the `wings` definition in your JSON schema with this:

```
        "wings": {
          "type": "object",
          "properties": {
            "type": {
              "allOf": [
                { "type": "string", "minLength": 1, "enum": [ "monocoque", "braced" ] }
              ]
            }
          },
          "required": [
            "type"
          ]
        },
```

tdakanalis:
Sorry for my delayed reply. I looked at the fix and it seems OK.

maaikez:
Thanks for fixing it so quickly.

I'm sorry for not being clear about the 'allOf' issue. I am using refs in the allOf arrays. So what I actually mean is: it doesn't work with refs.
"$ref": "#/definitions/wing"

josdejong:
@tdakanalis thanks for the feedback.

@maaikez I get it. The script indeed doesn't yet resolve paths of `$ref`. I've created a new issue for this: #302. I will do a bug fix release right now with the latest fixes.

PR #288, opened 2016-04-01
Add option to sort objects keys alphabetically

edufelipe:
The new option is disabled by default, but when enabled makes editing large documents in tree or form mode much easier.

It also includes the nicety of being a natural sort, so object that have numbers inside strings as keys sort very well.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -68,6 +68,10 @@ Constructs a new JSONEditor.
 
   If true, unicode characters are escaped and displayed as their hexadecimal code (like `\u260E`) instead of of the character itself (like `☎`). False by default.
 
+- `{boolean} sortObjectKeys`
+
+  If true, object keys in 'tree', 'view' or 'form' mode list be listed alphabetically instead by their insertion order. Sorting is performed using a natural sort algorithm, which makes it easier to see objects that have string numbers as keys. False by default.
+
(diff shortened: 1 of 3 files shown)


josdejong:
That's a nice idea. Thanks for your implementation including docs and everything, it looks neat!

josdejong:
@edufelipe I applied natural sorting to the sorting via the context menu too for consistency. I needed a comparator function `naturalSort(a, b)` for this though instead of the `naturalSort(array)` method that you provided, I included the [javascript-natural-sort](https://www.npmjs.com/package/javascript-natural-sort) library for this to save me the work of reinventing the wheel. See: https://github.com/josdejong/jsoneditor/commit/e7b7362fc2cd10ad3be78dee805b831b497cb9df

edufelipe:
@josdejong You know what's funny? I knew about that library and did not use it on purpose, mainly because your project is not heavy on dependencies and `left-pad` just happened. I was sure you would reject my pull request if I included another library. Oh well, that's what I get for not asking about it first. 😄

josdejong:
ha ha :)

I guess you agree with the current approach?

edufelipe:
Sure! As long as there is an option to do it programmatically on JSON load, everything works for me.

PR #300, opened 2016-05-17
-Added the ability to distinguish among nodes with enum type and the …

tdakanalis:
…same name. The schema property "id" must be defined in those nodes in a form that will identify the exact path of the node. For example:

"type": {
  "id": "airplanes/motor/type",
  "type": "string",
  "minLength": 1,
  "enum": [ "jet a1", "avgas" ]
}


josdejong:
Thanks. I've implemented a different fix which doesn't require an `id`, See https://github.com/josdejong/jsoneditor/commit/b922aceb07c4a714aaa310af6a281f35b7b9db8a.

PR #304, opened 2016-06-03
Added save dialog to example 4

RDCH106:
Added dialog asking for the file name to save it

### DIFF
--- a/examples/04_load_and_save.html
+++ b/examples/04_load_and_save.html
@@ -52,8 +52,21 @@
 
   // Save a JSON document
   document.getElementById('saveDocument').onclick = function () {
+    // Save Dialog
+    fname = window.prompt("Save as...");
+    
+    // Check json extension in file name
+    if(fname.indexOf(".")==-1){
+      fname = fname + ".json";
+    }else{
+      if(fname.split('.').pop().toLowerCase() == "json"){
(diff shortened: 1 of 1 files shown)


josdejong:
Nice improvement, thanks Ruben.

Two little things:
- can you please redo your PR to the `develop` branch instead? I can't accept it on `master`
- I'm not sure if it's needed to take special care for a `.json` suffix in the file name: if people want to name there file myDocumet.txt or whatever that's fine I guess, and this extra code distracts from the key point of the example. What do you think?

RDCH106:
- No problem. I will create a pull request to develop branch.

- The extensions are orientative. Usually the extensions in the operative systems are to asign an icon and default program. I think that it has not sense to create a standard json file with different extension to json. Not include extension is an other option, but it is more informative with the extension. About the clearity of the code, the code includes comments and the save option is full functionally. Ready to use! In my opinion is better than without extension checking.

Have I your aprovement to create a new PR in develop?

josdejong:
Ok thanks for your feedback. Let's keep the extension handling code in, it's fine with me. I will await your PR on develop.

RDCH106:
OK

josdejong:
I will close this PR now in favor of #305

PR #316, opened 2016-07-12
Add a callback for object ID labels

nucleartide:
Hey @josdejong! So at work I had a need to add ID labels for object nodes in the tree mode editor. Here's my implementation. What do you think?

(Had to make a new PR since I renamed the branch on my fork.)


josdejong:
Hi Jason, interesting, thanks for sharing your solution. I think your ID labels are just one case where you want to customize the way JSONEditor renders fields and values. I think the better solution is to implement life-cycle hooks where you can adjust these sort of things without having to adjust JSONEditor itself.

nucleartide:
Ah, good idea - I'll have a second look at this at work if I get the chance.

nucleartide:
Going to close this. It seems like this will be addressed by #337. In the meantime, I can continue to depend on the JSONEditor fork.

Issue #314, opened 2016-07-04
JSON parse error when displaying validation error in tree mode


powerslave:
Using schema and json provided below, I get a JSON parse error, and a broken tree view.
`Failed to parse JSON schema: SyntaxError: Unexpected token ' in JSON at position 0`

This schema/json combo is supposed to fail validation. I verified proper failing validation with both http://jsonschemalint.com/draft4/ and http://www.jsonschemavalidator.net/ The failing part in json is `"outside":{}`

* I can repro this both on `jsononlineeditor.org` and a local app that loads `cdnjs/5.5.6/jsoneditor.min.js`
* Issue seems to happen when when _displaying_ the validation error, validation per se works fine.
* Only happens in tree view. Validation error display in code view is correct.

##### reproduction:

1: set this schema: http://www.jsoneditoronline.org/?id=92b1743f2e0d71f85b406818f0538841

```json
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "definitions": {
    "schemaArray": {
      "type": "array",
      "items": {
        "$ref": "#"
      }
    }
  },
  "type": "object",
  "properties": {
    "$ref": {
      "type": "string"
    },    
    "additionalProperties": {},
    "definitions": {
      "type": "object",
      "additionalProperties": {
        "$ref": "#"
      }
    },
    "properties": {
      "type": "object",
      "additionalProperties": {
        "$ref": "#"
      }
    },
    "type": {}
  },
  "additionalProperties": false
}
```

----------------

2: and load this json: http://www.jsoneditoronline.org/?id=6a1361edd42146d7c93c0cb9f746f2b8

```json
{
  "type": "object",
  "properties": {
    "Sub": {
      "type": "object",
      "$ref": "#/definitions/sub.json"
    }
  },
  "definitions": {
    "sub.json": {
      "type": "object",
      "properties": {},
      "additionalProperties": false,
      "outside":{}
    }
  }
}
```

Happy to help with more details

josdejong:
Thanks for reporting, this issue originates with JSONEditor not yet understanding `$ref` in a schema.

See also #302 and #310

powerslave:
thanks @josdejong 
FYI: In one of my scenarios I could workaround by adding `onEditable: noEditing`, keeping validation alive in tree view. In the other scenario `noEditing` did not help, have to live with validation in text view only for now.

josdejong:
That makes sense, thanks for the feedback.

[...]

PR #326, opened 2016-09-22
Fix https://github.com/josdejong/jsoneditor/issues/314

apostrophest:
The JSON path parsing code assumes that string keys within square brackets in JSON paths will be valid, parseable JSON.

However, the ajv library on which this project depends emits string keys in single quotes.

For example, ajv can emit .foo['bar'], which will ultimately end up with a call to JSON.parse("'bar'"). This causes the "Unexpected token ' in JSON at position 0" error thrown from JSON.parse.

 Since ajv shouldn't need to change and since this utility is gone in the next branch, this fix checks for ajv-formatted string keys and patches them to double-quoted strings before calling JSON.parse. This is the narrowest fix for the problem.


apostrophest:
Didn't mean to include https://github.com/josdejong/jsoneditor/pull/326/commits/fe459e64c37af0748af2b4e02ebdab3a0e2badaa in this PR. Let me know if you want me to submit a cleaner PR.

josdejong:
Thanks for the fix Stephen!

I normally only update build files (under /dist) when doing a release, but it's no big deal, will merge your PR now.

josdejong:
I've release v5.5.8 containing your fix Stephen.

PR #338, opened 2016-11-22
Customize object names

chromey:


joneal:
+1

This would be a huge UX improvement.

trsh:
+1

johanbm1405-commits (approved):

PR #355, opened 2017-01-18
Suggesting a different directive for Angular.js.

isonet:
Hey, first of all I would like to thank you for this awesome tool you built.

I tried using the referenced angular directive, but it was very out of date and difficult to use...

So I wrote my own wrapper and thought you might want to add it to your project's readme.
The new one works with the latest version of jsoneditor.

Thanks in advance!

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ with bower:
 
 There is a directive available for using JSONEditor in Angular.js:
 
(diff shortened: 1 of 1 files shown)


josdejong:
:+1: 

You could try to get your fixed merged into the original directive by sending a PR, that would be event better.

PR #381, opened 2017-03-30
Optimize the input experience of Chinese IME

chinesedfan:
By default, ace accepts user inputs by textarea.ace_text-input, which has a cursor. If typing with IME, another cursor appears.

In the next picture, line 11 has the first cursor at the right of "a", while the second one is after "wen"(it is flashing, not very clear).

4 pic

In fact, ace tries to avoid this situation by setting z-index of class ace_composition, which is added to textarea.ace_text-input in composition input mode. But its background color is "inherit". If the parent element's background color is transparent, the first cursor can not be covered.

Official themes of ace always set background color of the parent element. So maybe our theme-jsoneditor should keep the same. After fixed, the input experience is better.

3 pic 1

ps. I think minor PRs can be send to master directly. But if you persist, I will fix them.

### DIFF
--- a/src/js/ace/theme-jsoneditor.js
+++ b/src/js/ace/theme-jsoneditor.js
@@ -40,6 +40,7 @@ color: #333\
 .ace-jsoneditor.ace_editor {\
 font-family: droid sans mono, consolas, monospace, courier new, courier, sans-serif;\
 line-height: 1.3;\
+background-color: #fff;\
 }\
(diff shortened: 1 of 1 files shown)


josdejong:
Thanks Xianming!

PR's should indeed go to the `develop` branch. When doing a release, `develop` is merged in `master`.

chinesedfan:
Sure. I will remember this to lighten your workload. :)

Issue #382, opened 2017-04-08
Readonly text mode.


walkerrandolphsmith:
Hi @josdejong. I see that when the editor is in the tree mode I can decide which nodes are editable with the onEditable option. If I need all the nodes in the tree to not be editable I can use the view mode. I want to Use the text mode to display a read only view of the raw text. Textareas support a readonly attribute. Is there a setting I have overlooked or would this be a valid feature to implement. I can submit a PR if you like.

josdejong:
Fixed in v5.6.0 with your PR #383. Thanks again :)

PR #383, opened 2017-04-08
#382 Read only text mode.

walkerrandolphsmith:
Addresses [#382](https://github.com/josdejong/jsoneditor/issues/382). I have also added an example file in addition to the changes in `./src`. I didn't want to make the surface area of `options` any larger, so I decided to use `onEditable` to determine if the textarea in `text` mode is editable. If the `onEditable` key is absent in the `options` than the content is editable. If the key is present and has a value that is a function, it will determine whether the text area is editable based on the result of `onEditable`'s invocation.

If hijacking the `onEditable` option is not appropriate I'd love to chat about the desired api of `options`.

### DIFF
--- a/examples/06_readonly_text_mode.html
+++ b/examples/06_readonly_text_mode.html
@@ -1,0 +1,69 @@
+<!DOCTYPE HTML>
+<html>
+<head>
+  <title>JSONEditor | Switch mode</title>
+
+  <!-- when using the mode "code", it's important to specify charset utf-8 -->
+  <meta http-equiv="Content-Type" content="text/html;charset=utf-8">
+
+  <link href="../dist/jsoneditor.css" rel="stylesheet" type="text/css">
+  <script src="../dist/jsoneditor.js"></script>
+
+  <style type="text/css">
+    body {
+      font: 10.5pt arial;
+      color: #4d4d4d;
+      line-height: 150%;
+      width: 500px;
+    }
+
+    code {
+      background-color: #f5f5f5;
+    }
+
+    #jsoneditor {
+      width: 500px;
+      height: 500px;
+    }
+  </style>
+</head>
+<body>
+
+<p>
+  Switch editor mode using the mode box.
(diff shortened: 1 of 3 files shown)


josdejong:
Thanks Walker, looks good!

Two remarks:
- It would be nice if we could make the editor readonly when in mode `code` too
- Thanks for adding an example too. Can you rename it to `09_readonly_text_mode.html`?

josdejong:
> If hijacking the onEditable option is not appropriate I'd love to chat about the desired api of options.

I agree with you, using `onEditable` for making the text mode readonly feels a bit like misusing the existing option. On the other hand it keeps the API compact and consistent. I think your solution is elegant.

walkerrandolphsmith:
Okay, I can take a look into supporting read only `code` modes. I can rename the example file. I am surprised I didn't notice the convention!

walkerrandolphsmith:
So that should allow the `code` mode to be configured to be read only as well.

There was a moment where I could't decide which styling was most appropriate:

```
aceEditor.setOptions({ readOnly: isReadOnly });
```

```
aceEditor.setOptions({readOnly: isReadOnly});
```

```
aceEditor.setOptions({
    readOnly: isReadOnly
});
```

Would you be interested in adding lint rules and [eslint](http://eslint.org/) to enforce style guides outlined in the [contributing](https://github.com/josdejong/jsoneditor/blob/master/CONTRIBUTING.md) doc? I can spin off a new feature and associated PR.

walkerrandolphsmith:
Rebased the upstream develop branch.

josdejong:
Thanks, looks good! 

Reminder to myself: add a sentence about that you can use `onEditable` for modes `text` and `code` too in the docs.

josdejong:
As for the styling: I tend to use the styling you used in the PR too lately. A linter would be welcome! Would be nice if you could set that up that for jsoneditor, thanks beforehand.

PR #396, opened 2017-05-05
Feature/template items

israelito3000:
Hello, These changes allow user to declare templates in order to add them to the menu just like #75, Templates example is provided as well, I was thinking in implementing by populate the menu using the json schema, but still I think this could be very valuable for non schemed jsons.

Template are declare in options like that:

```
var options = {
      templates: [
          {
              "menu": "Special Node",
              "name": "SpecialNode",
              "class": "my-special-css-class",
              "data": {
                  "field1": "value1",
                  "field2": "value2"
              }
          }
      ]
  };
```

data contain the node(template) that will be inserted/appended.

Regards.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,3 @@ downloads
 node_modules
 *.zip
 npm-debug.log
-/.vs
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ downloads
 node_modules
 *.zip
 npm-debug.log
+/.vs
--- a/examples/10_templates.html
+++ b/examples/10_templates.html
@@ -1,0 +1,59 @@
+<!DOCTYPE HTML>
+<html>
+<head>
+  <title>JSONEditor | JSON schema validation</title>
+
+  <link href="../dist/jsoneditor.css" rel="stylesheet" type="text/css">
+  <script src="../dist/jsoneditor.js"></script>
+
+  <style type="text/css">
+    body {
+      width: 600px;
+      font: 11pt sans-serif;
+    }
+    #jsoneditor {
+      width: 100%;
+      height: 500px;
+    }
+
+  </style>
+</head>
+<body>
+<h1>JSON schema validation</h1>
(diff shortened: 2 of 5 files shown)


josdejong:
That's a nice solution! Thanks Israel. If you don't mind I will work out the example a bit more (and add docs).

josdejong:
Thinking about it, how about renaming the properties of the templates to:

```js
var options = {
      templates: [
          {
              text: 'Special',
              title: 'Insert a Special Node'
              className: 'my-special-css-class',
              field: 'SpecialNode',
              value: {
                  field1: 'value1',
                  field2: 'value2'
              }
          }
      ]
  };
```

That better matches the current API (which uses `field` and `value`), and allows to enter a short menu `text` and a longer `title`.

israelito3000:
I think is a good idea, do you want me to work on it and create a pull req?
By the way, I was thinking in how to include autocomplete feature like Ace editor, any suggestion how to approach?
Regards

josdejong:
If you want to create a new PR would be great but I can do it too if you're not able to work on it.

Let's discuss your suggestion for autocompletion in it's own topic, I created an issue: #398

israelito3000:
Check this out: https://israelito3000.github.io/jsoneditor/11_autocomplete.html

Tested in Chrome.

The autocomplete engine are external, so is not need to be embedded into the component, just need to have autocomplete option with the following interface:

autocomplete : {
    ActivationChar : '',   // This is a character that makes the autocomplete activate if the line startswith it.
    ApplyTo:['values']    // it can be 'values' or 'name' depending where you want autocomplete feature.
    Show: function (node, element){
        // Show the autocomplete 
     }
    Hide: function() {
       // hide the autocomplete
    }
}

In the example I create autocomplete options for create jsonpaths based on json structure, you can see more details in the branch: https://github.com/israelito3000/jsoneditor/tree/feature/autocomplete

What do you think? does this feature can apply for a pull request?

Regards.
Israel.

Issue #398, opened 2017-05-09
Autocompletion


josdejong:
In https://github.com/josdejong/jsoneditor/pull/396#issuecomment-299991903, @israelito3000 suggests implementing autocompletion. 

Not sure how that may best work, maybe something like collect all words in the JSON document and create a dropdown with autocompletion options whilst typing.

israelito3000:
Check this out: https://israelito3000.github.io/jsoneditor/11_autocomplete.html

Tested in Chrome.

The autocomplete engine are external, so is not need to be embedded into the component, just need to have autocomplete option with the following interface:

```
autocomplete : {
    ActivationChar : '',   // This is a character that makes the autocomplete activate if the line startswith it.
    ApplyTo:['values']    // it can be 'values' or 'name' depending where you want autocomplete feature.
    Show: function (node, element){
        // Show the autocomplete 
     }
    Hide: function() {
       // hide the autocomplete
    }
}
```

In the example I create autocomplete options for create jsonpaths based on json structure, you can see more details in the branch: https://github.com/israelito3000/jsoneditor/tree/feature/autocomplete

What do you think? does this feature can apply for a pull request?

Regards.
Israel.

josdejong:
That's cool, your example works like a charm!

It may be a good idea to keep the implementation of the autocomplete engine externally, though I'm not sure whether such an API (where people have to hook up their own solution) works nicely with existing autocompletion libraries. Do you know whether that's the case? If not, it may be better to implement the rendering of this dropdown ourselves, and only expose an `onAutocomplete(text)` function which allows you to return an autocompleted version of the input text.

israelito3000:
Hello Jos.

Thank you for your feedback, actually I was researching before work on autocomplete and the main issue is that most of the autocomplete libraries rely on input text fields, which is not ahow jsoneditor works, I found one "horsey autocomplete" and I integrated without too many issues, I actually set the example here:

https://israelito3000.github.io/jsoneditor/12_autocomplete_horsey.html

It need to be refined a little bit to accept "." as autocompletion trigger, to work similar to the one I did .. but this is also specific for my scenario of jsonpath options. 

If you dont mind, I am going to create a more easy example to only display a set of options base on the node content, not using the jsonpath validation which is very specific for a project that I am working on. and I am going to create a pull request. 

There is not too many changes in the core of jsonpath to achieve this feature actually.

josdejong:
[...]

PR #407, opened 2017-05-26
Feature/autocomplete2

israelito3000:
Hello Jos,

I integrate the autocomplete and include some examples about how works. Tested with IE 11 and chrome.

You can see examples working at:
https://israelito3000.github.io/jsoneditor/11_autocomplete.html
https://israelito3000.github.io/jsoneditor/12_autocomplete_simple.html
https://israelito3000.github.io/jsoneditor/13_autocomplete_simplest.html

Regards.
Israel.


josdejong:
Thanks for working this out @israelito3000 ! I like the "simplest" example :). Maybe the first example should be the "simplest", and a second example should be named "advanced" or something.

I will review your code more in-depth soon and do functional testing. Want to think the API through. A few small remarks already:

- I think the styling should go into a CSS file
- method and property names should be lowerCamelCase to be consistent with the rest of the code (for example `GetOptions`, `Show`, `ActivationChar`.

related: #398

israelito3000:
Hi Jos
Thank you for your observations, you are totally right, I apologize, I adjusted the code, refined the examples and include the CSS styles according to the project standard structure.

All examples here:
https://israelito3000.github.io/jsoneditor/

Let me know anything else.

Regards.
Israel.

josdejong:
I've finally thoroughly checked your work, sorry for the delays. Anyway, I'm really looking forward to see autocompletion support in the editor :)

Here quite some feedback (mostly centered around trying to achieve a concise and consistent API and user experience). Thanks for reworking the examples, that's already more approachable like this.

- First, it works like a charm, great job!

- Code

  - Your code looks neat, little to add to that
  - I still see a lot of styling in `autocomplete.js` (search for occurences of `.style`, can you move that to a separate CSS file? Or is that with a specific reason? 

  - Inside `autocomplete.js` you extend a number of native objects with methods `remove` and `startsWith`. Are these methods conform ES6, are they proper polyfills? You could optionally move them to the `util.js` file where there are more polyfills like these (for example `addEventListener` and `removeEventListener`).
  - Most functions are now lower camel case, only the method `Show` still starts with upper case, would be neat to make that lower case.

  - Most code in the two advanced examples is code for generating a list with options from the current JSON document (extract all words). I think the examples will become easier to understand when this code is extracted and put at the bottom of the example, and maybe rename it to `function extractUniqueWords (json)` or something. Since this code is not the core of the example, it may be interesting too to simplify it by loading lodash in the example and use flatMapDeep, something like:

    ```js
    <script src="https://cdnjs.cloudflare.com/ajax/libs/lodash.js/4.17.4/lodash.min.js"></script>
  
    // ...
    
    function extractUniqueWords (json) {
      return _.uniq(_.flatMapDeep(json, function (value, key) {
        return _.isObject(value) 
            ? [key] 
            : [key, String(value)]
      }))
    }
    ```

- Styling improvements:
  - Same styling as the context menu: same border color and shadows.
  - Hide scrollbar when it's not needed.

- Docs
  - Can you please write docs on the `/docs/api.md` page? I realize we also forgot to write docs for the `template` functionality...

- API

  - The solution currently assumes one word will be entered in a field, it does not work for a field containing a complete sentence. I think that's fine. When there is demand for a more advanced version we can always extend it.

  - Currently only support for synchronously getting options. I think that's fine: the current method `getOptions` can be extended in the future to allow returning a Promise.

  - It may be better to rename the option values of `applyTo` to `field` and `value` instead of `name` and `value` to be consistent with other options of the editor.

  - I'm reserved about exposing the internal nodes as second argument of `getOptions`. So far internal nodes are not exposed via the API because the internal API of the nodes is not suitable for external use (I want to address that in v6). Instead of exposing the nodes itself, it's probably enough to expose the path and whether the field or value of the node has focus, similar to how the `onEditable` callback works. 

  - I'm not sure whether `activationChar` is useful in practice, do you have use cases for this? If there are no clear/common use cases I would prefer to leave it out to get a simpler API.

  - I think the `confirmKeys` can be configured directly inside `autocomplete` instead of nested in `autocomplete.config`. We are already inside configuration, and configuration of configuration sounds a bit duplicate :D

- Questions

  - How to apply different set of options for either a field or a value of the same node? Is that possible right now? Maybe we can pass `"field"` / `"value"` as an argument to the `getOptions` function instead, so you can dynamically decide there?

israelito3000:
Hello Jos, I made an adjustment of most of your suggestions, please check the commit comments, here my comments:

*"I still see a lot of styling in autocomplete.js (search for occurences of .style, can you move that to a separate CSS file? Or is that with a specific reason?*"

**R://** The stylings remaining in "autocomplete.js" are related to positioning the scrollbar and the hint and for calculate the size of the controls based on the content, but not for really styling.


*"The solution currently assumes one word will be entered in a field, it does not work for a field containing a complete sentence. I think that's fine. When there is demand for a more advanced version we can always extend it."*

**R://** Lets do it for a next version.

*"Currently only support for synchronously getting options. I think that's fine: the current method getOptions can be extended in the future to allow returning a Promise."*

**R://** I implemented by allowing both cases. Let me know if this is ok for you.

*"I'm reserved about exposing the internal nodes as second argument of getOptions. So far internal nodes are not exposed via the API because the internal API of the nodes is not suitable for external use (I want to address that in v6). Instead of exposing the nodes itself, it's probably enough to expose the path and whether the field or value of the node has focus, similar to how the onEditable callback works."*

**R://** I had a very specific scenario in which I needed the node, I needed to look for a node by jsonpath, in the advanced demo I called: `node.editor.node.findNode`, maybe you can suggest me some other way to achieve this?

*"I'm not sure whether activationChar is useful in practice, do you have use cases for this? If there are no clear/common use cases I would prefer to leave it out to get a simpler API."*

**R://** In my scenario I only want to launch the autocomplete when the text starts with an specific character, I can remove this and make the check in the function instead. what do u think?

josdejong:
Thanks! I will check out your changes and comments asap but need some time for it, hopefully coming Tuesday evening.

israelito3000:
Thank you.
Israel.

josdejong:
I like the improvements you did, and also the docs on the new features that you've worked out. Thanks.

Here some more feedback. I think we're getting there but there is still some polisihing needed for the API. I've numbered the remarks for easier discussion.

1. You have an interesting take on CSS, wanting to limit it to things related to (customizable) styling. I always find CSS a bit confusing exactly because half of it is related to positioning and the other half to styling. Let's just keep it this way, it's ok with me :).

2. Both sync and async support is just great! 

3. Good idea to move the `activationChar` functionality to the `getOptions` callback! So for example this callback returns `null` or `undefined` instead of an `Array`, it simply means that we don't want to do autocompletion, and the user has all freedom to decide.

4. It took me a while before I understood what the advanced example (13) exactly does. It's indeed quite an advanced example, which is great - so we can see how flexible the API is. We should try to separate special requirements for this advanced example from coupling with internals of JSONeditor. I think that it's possible not to use internal `node.editor.node.findNode()`: In the callback you can call the public `editor.get()` to get the full JSON, and then use a [jsonpath](https://www.npmjs.com/package/jsonpath) library to see where what the possible paths are from the current path. There is one other very tricky think in the example: `autocomplete.startFrom` is manipulated. If I'm not mistaken this is to prevent overriding the already entered part of the path. My feeling though is that this is only needed because the autocompletion doesn't really support autocompletion of multiple words or parts of the input. Maybe we should address that instead.

Just thinking aloud about the exact API here, let's try to converge. How about something like:

```js
autocomplete: {
    confirmKeys: [39, 35, 9, 190],
    getOptions: function (text, path, input, caret) {
        // text is the text that's entered so far
        // path is a an array with strings of the current path (same as onEditable)
        // input is the name of the current input: 'field' or 'value'
        // caret is an object {start: number, end: number} telling where 
        // the caret currently is
        
        // return null for no autocompletion
        return null;

        // return an array with replacements which will fully replace text
        return ['field1', 'field2'];    

        // return an array with objects to allow replacing a part of the text, allowing
        // to replace one of multiple words or parts. Start and end are both optional.
        // for example when text is 'hello wo' and the caret is at the end, an option 
        //     [ { start: 6, replacement: 'world' } ]
        // to replace 'wo' with 'world'
        return [
            {start: 8, end: 12, replacement: '.field1'},
            {start: 8, end: 12, replacement: '.field2'}
        ]

        // return a Promise for async support
    }
}
```

Some thoughts behind this API:

- Basic usage with a static list is simple
- No need for options like `activationChar` or `applyTo`, user can decide in the `getOptions` callback based on the `path` and `input` arguments.
- No need to expose internals like node, editor, findNode, or autocomplete instance
- support for autocompletion of multiple words/parts

What do you think?

israelito3000:
I liked your suggestions, I really appreciate your time and dedication to do code review, thank you very much for that.

I improved the getOptions by remove the node and autocomplete from the parameters, achieving jsonpath query by using an external library.

I also removed applyTo options letting user manage that from getOptions like your suggestion, more clean!

I also update the documentation and samples according to the changes.

About multiple replacement I think by returning an object like: {start:4, options:["orld","almart"]} could work, instead of returning multiples start and end.. actually I couldn't see how you figure that.

For multiple replacement "caret" parameter still need to be implemented, If you agree, I prefer implement that feature in a next version.

josdejong:
awesome Israel! Love the final result :+1: 

I agree with not implementing support for autocompleting words halfway a sentence right now, but it's good to know that the current API has the flexibility to extend it in the future.

josdejong:
I've released v5.8.0 containing autocompletion support :)

josdejong:
@israelito3000 a user just reported me via email about two issues he came across with the new autocomplete at the place where `getOptions` is called:

```js
var result = this.options.autocomplete.getOptions(element.innerText, editor.get(), jsonElementType);
```

-  the code should not call a global `editor.get()` but `this.get()`
- the way `getOptions` is not called as `getOptions(text, path, input)`: the second argument currently still is the complete JSON object instead of the current path.

I should have seen this during the review :flushed: . Can you have a look at this?

israelito3000:
Hello Jos

I am going to take a look.

israelito3000:
Jos,
I don't see the issue, inside getOptions we need the whole json object in order to extract all the words (example dynamic) maybe the name of the parameter "path" is not ok, but if I send only the current path, how can I analyse the whole document?

josdejong:
You can call `editor.get()` yourself inside `getOptions` if needed like in the dynamic example. It's not needed to pass the object as argument of `getOptions`. 

The `path` however is important to pass as argument, you cannot determine that from the outside. The path is needed in use-cases where your for example want to autocomplete a field like `.user.address.city` with names of all cities. Does that make sense?

israelito3000:
Ok, I implemented by your suggestions, I am going to submit another pull request.
Just one suggestion: I pass "editor" object as the first parameter in getOptions, I believe is a good practice when events send the "sender" and is better to use this local variable instead of the editor global variable, not always users have globals available, and in case of more than one editor could be an issue working with globals. Do you agree?

josdejong:
Thanks!

Very good idea to also pass the `editor` instance as argument. This is also needed for the `onChange` callback since it's possible that this callback is invoked whilst constructing the editor, see #420 . I think though that we can better pass it as last argument since it's only sometimes needed whilst `text` is teh most important argument and should go first. Similar to the JavasScript API's of `map` and `filter`, which pass the Array as third and last argument.

As for the path itself: I think we should change that to returning an Array with strings like `['user', 'address', 'city']` instead of a JSONPath like `$.user.address.city`. That will keep it consistent with the API of the `onEditable` callback which also passes an array with strings. We can argue about what would be the best to pass as `path` (JSONPath or Array), but let's keep the API consistent and discuss this a separate discussion if you it would be worth changing the API in this regard.

josdejong:
Just for reference: we're discussing #426 above

Issue #302, opened 2016-05-24
JSON schema enums not resolved via $ref in the schema 


josdejong:
The editor doesn't create a dropdown for enums in a JSON schema which have to be resolved via a reference `$ref` like in:

```
{
    "type": "array",
    "items": { "$ref": "#/definitions/myEnum" },
    "definitions": {
        "myEnum": {
            "enum": ["one", "two", "three"]
        }
    }
}`
```

See also https://github.com/josdejong/jsoneditor/pull/285#issuecomment-221183267

maaikez:
Thanks for adding this issue. What is the priority of this feature?

josdejong:
It's not a bug so there is no extremely high priority on this. If you find time to implement this before I do, feel free to contribute :).

maaikez:
That was indeed what I thought. There is no time for the project I used the json editor for anymore but if I can do it at home, I will try to help (but that's not very soon).

zgz682000:
but why "$ref" works on https://www.jsoneditoronline.org/

josdejong:
The JSON schema error messages work but there is no enum dropdown created

josdejong:
Fixed via #423

PR #423, opened 2017-07-04
add support for json schema references ($ref)

meirotstein:
Add support for JSON schema validation for schemas which has references to other schemas using the '$ref' option.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -228,7 +228,7 @@ Set a field name for the root node.
 
   Field name of the root node. If undefined, the current name will be removed.
 
-#### `JSONEditor.setSchema(schema)`
+#### `JSONEditor.setSchema(schema [,schemaRefs])`
 
 Set a JSON schema for validation of the JSON object. See also option `schema`.
 See [http://json-schema.org/](http://json-schema.org/) for more information on the JSON schema definition.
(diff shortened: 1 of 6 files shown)


josdejong:
Thanks Meir for your PR! I will look into it coming Friday or Saturday.

josdejong:
Thanks again, your code looks neat @meirotstein ! This is a nice new feature. Just for reference: this PR solves #302.

One thought: instead of a new method `setSchemaRef`, we could simply pass the schema refs as an optional second argument of the existing method `setSchema`, like `setSchema(schema [, schemaRefs])`. The `schemaRefs` argument would be exactly the same as the new option `schemaRefs` (an object with ref as key, schema as value), and this way both schema and refs are applied in a single action instead of two consecutive actions. What do you think?

meirotstein:
Hi - thanks for the feedback!

I agree - your suggestion is much more cleaner then adding a new method - I will push an update later on today.

josdejong:
Thanks!

josdejong:
Looks good, thanks again @meirotstein !

josdejong:
JSON schema refs is now available in `v5.9.0` :tada:

meirotstein:
Awesome!
Thank you very much!

PR #424, opened 2017-07-05
Enabled the default mode to be set when using a list of available modes.

bag-man:
I noticed that when using a list of available modes say `[ 'code', 'form', 'tree' ]`, that `tree` would always be selected as the default, when it would be nice to be able to have the first mode in the list used as the default. 

Upon further inspection I also realised that if `tree` wasn't in the list it would still be shown as the default, and then not made available as an option to change to, so the user would only see it on first load. 

This simple change fixes these issues.


josdejong:
Thanks Owen, nice fix!

PR #426, opened 2017-07-07
Improvements/autocomplete

israelito3000:
Improvements to autocomplete to pass editor and jsonpath in getOptions method.


israelito3000:
Good, I changed the order of options to the last position, and change the path (sending now standard array path instead of jsonpath as you suggest in: #407).

josdejong:
Thanks for your effort Israel, looks good!

josdejong:
Your improvements are now available in `v5.8.2`

PR #428, opened 2017-07-10
Add extra verification when creating JSON editor to avoid to look for…

trystan2k:
… an element in an empty array


trystan2k:
Related with issue (https://github.com/josdejong/jsoneditor/issues/429)

josdejong:
Thanks! That's quite a special case...

I now see that we have to change the order of condition too: first select `this.options.mode`, else use the first entry from this.options.mode, else use 'tree'. Will adjust that too.

Issue #435, opened 2017-07-17
Custom button


zendre4:
Hello,

Have you a solution to add a custom button ? I woul'd like add a button in the toolbar, for copy the JSON (with https://clipboardjs.com/)

Thanks

dhughes-xumak:
Similarly, it would be great if the context menu also supported plugin actions.

Both menu bar and context menu could be extensible through new options in the JSONEditor constructor.

dhughes-xumak:
I haven't tested it yet, but here is a stab at supporting toolbar plugins: https://github.com/dhughes-xumak/jsoneditor/commit/c55ee1cf0d1d3e47d03534c160f72fefbd7c828a

Context menu plugins are pending.

Issue #337, opened 2016-11-07
Progress on version `next`


josdejong:
I'm working on a completely new version of the JSONEditor. The current code base is hard to maintain and expand, and is too limited qua customization. This new version aims at making the editor extendable (i.e. customize the action menu, custom modes, etc). The new editor will have good support for synchronizing changes with other editors or with a backend via JSON-Patch actions. 

### Try it out now

The new editor is being built in the [next](https://github.com/josdejong/jsoneditor/tree/next) branch. To try it out:

- clone the project: `git clone https://github.com/josdejong/jsoneditor.git`
- go to the project directory: `cd jsoneditor`
- check out the `next` branch with git: `git checkout next`
- install dependencies: `npm install`
- build the library: `npm run build`

All examples in the `examples` folder are basically working and be tried out.

### Progress of version `next`

- [x] Set up structure for the new editor, using react or preact
- [x] Implement a new data model (works with JSON-Patch to apply changes)
- [x] Implement method `patch` and an `onChange` listener using JSON-Patch
- [x] Implement modes Tree, Form, View 
- [x] Implement modes Text and Code (the latter integrating Ace)
- [x] Implement mode switcher and API
- [x] Integrate jsonlint and handle errors nicely
- [x] Implement configurable readonly properties and values
- [x] Implement history, undo, redo
- [x] Implement JSONSchema
- [x] Implement search
- [x] Implement customizable quick keys
- [ ] Implement drag and drop of items
- [ ] Implement checkbox for booleans
- [ ] Implement select box for enums (when using JSONSchema)
- [ ] Implement `patchText` and `onTextChange` when in modes Text or Code 
- [ ] Expose an API for React
- [ ] A bunch of small fixes, refactorings, and performance improvements

### Extra

- [ ] Implement hooks to customize the main menu
- [ ] Implement hooks to customize the action menu
- [ ] Implement hooks to customize how values are rendered (i.e. like show a checkbox for boolean values, select box for enums, etc)
- [ ] Implement API to register new modes
- [ ] Implement a more advanced Repair button than the current one that's implicitely executed when clicking format/compact in text mode. Should turn JS data into JSON, and help you to repair JSON-like data with a comma missing somewhere or things like that.

hassankhan:
Would it be possible to rewrite styles in SCSS for v6? Happy to help with a PR if so 😄

josdejong:
[...]

PR #438, opened 2017-07-18
Implement support for Toolbar and Context Menu plugins

dhughes-xumak:
Related to #435.  This adds options for easily defining additional buttons to the Toolbar and Context Menu.  Though multi-select is currently broken, this should support additional custom Context Menu options for multiple selections.

---
This contribution is graciously provided by **XumaK** (http://xumak.com/)

_XumaK helps reduce the uncertainty and risk that is normally involved for companies trying to transform themselves digitally._  

_We provide professional services teams with years of experience to deliver the project, using our end-to-end delivery platform, which ensures your project will be on time and on budget. We deliver world class customer solutions through innovative industry leading products._

### DIFF
--- a/src/js/JSONEditor.js
+++ b/src/js/JSONEditor.js
@@ -48,6 +48,16 @@ var util = require('./util');
  *                                                          buttons.  Must contain
  *                                                          `title`, `className`, and
  *                                                          `onclick` properties.
+ *                               {Object[]} contextMenuPlugins  Array of custom toolbar
+ *                                                              buttons.  Must contain
+ *                                                              'text', `title`, `className`,
+ *                                                              and either `_click` or
+ *                                                              `submenu` properties.
+ *                               {Object[]} multiContextMenuPlugins     Array of custom toolbar
+ *                                                                      buttons.  Must contain
+ *                                                                      'text', `title`, `className`,
+ *                                                                      and either `_click` or
+ *                                                                      `submenu` properties.
  * @param {Object | undefined} json JSON object
  */
 function JSONEditor (container, options, json) {
@@ -87,7 +97,7 @@ function JSONEditor (container, options, json) {
         'ace', 'theme','autocomplete',
         'onChange', 'onEditable', 'onError', 'onModeChange',
         'escapeUnicode', 'history', 'search', 'mode', 'modes', 'name', 'indentation', 'sortObjectKeys',
-        'toolbarPlugins'
+        'toolbarPlugins', 'contextMenuPlugins', 'multiContextMenuPlugins'
       ];
 
       Object.keys(options).forEach(function (option) {
--- a/src/js/JSONEditor.js
+++ b/src/js/JSONEditor.js
@@ -44,6 +44,10 @@ var util = require('./util');
  *                               {boolean} sortObjectKeys If true, object keys are
  *                                                        sorted before display.
  *                                                        false by default.
+ *                               {Object[]} toolbarPlugins  Array of custom toolbar
+ *                                                          buttons.  Must contain
+ *                                                          `title`, `className`, and
+ *                                                          `onclick` properties.
  * @param {Object | undefined} json JSON object
  */
 function JSONEditor (container, options, json) {
@@ -82,7 +86,8 @@ function JSONEditor (container, options, json) {
         'ajv', 'schema', 'schemaRefs','templates',
         'ace', 'theme','autocomplete',
         'onChange', 'onEditable', 'onError', 'onModeChange',
-        'escapeUnicode', 'history', 'search', 'mode', 'modes', 'name', 'indentation', 'sortObjectKeys'
+        'escapeUnicode', 'history', 'search', 'mode', 'modes', 'name', 'indentation', 'sortObjectKeys',
+        'toolbarPlugins'
       ];
 
       Object.keys(options).forEach(function (option) {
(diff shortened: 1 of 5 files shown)


josdejong:
Thanks a lot, appreciated! 

I will review your PR in-depth soon, it looks well taken care of at first sight.

dhughes-xumak:
Great, thanks @josdejong!  

We appreciate your quick action and on-going maintenance of this project.

dhughes-xumak:
Some other considerations to improve this feature:
* Add `{string} contextMenuPlugins.submenuTitle` property to the test data and documentation.
* Add `{function} contextMenuPlugins.enabled` property so each plugin can decide if it should be included in a given context menu.  This probably makes it possible to combine contextMenuPlugins and multiContextMenuPlugins into one list.
* Convert the built-in context menu actions to use the new plugin architecture (depends on the first two points).

Admittedly, my javascript is really rusty and I had some trouble with the context menu click handlers.  This is why they are being wrapped at https://github.com/josdejong/jsoneditor/pull/438/commits/2a6a65e889810551c591665d7eb01afd8d8afc53#diff-0e52f886d57f8a2c0f91f314276ac9a9R1295.  I'm sure the handlers could be constructed without needing to wrap them.
* Get rid of the callback wrapping for contextMenuPlugins/multiContextMenuPlugins.  If the selected node(s) are passed to the ContextMenu constructor, then all of the processing code could be moved to ContextMenu.

src/js/JSONEditor.js
```diff
@@ -44,6 +44,20 @@ var util = require('./util');
  *                               {boolean} sortObjectKeys If true, object keys are
  *                                                        sorted before display.
  *                                                        false by default.
+ *                               {Object[]} toolbarPlugins  Array of custom toolbar
```
dhughes-xumak: This docblock is outdated and should be removed entirely.  [API Doc](https://github.com/josdejong/jsoneditor/blob/develop/docs/api.md) should be sufficient for documenting this.

josdejong:
I've just checked out your PR @dhughes-xumak, thanks a lot again! You have implemented a solid API, you really know what you're doing and the code looks good.

On a side note: I'm working on a complete new version of the JSONEditor, "v6", see #337. I had planned features that you're implementing right now for v6.1. This means that we'll have to re-implement them some day for v6 (which is ok, most work is figuring out the right API, not the code itself).

Here some points of feedback:

1. For consistency, it may be better to rename the click callbacks in `toolbarPlugins`, `contextMenuPlugins`, and `multiContextMenuPlugins` all to `onClick`. Currently the first is `onclick` and the others are `click` following the documentation. So far the API is camelCase everywhere, so that's probably the most consistent naming to apply for onClick too.

2. Currently, the onClick callbacks pass the current node/nodes. The the Node class is non-public and fully undocumented, not really suitable API for public use. That will lead to questions and issues due to misunderstanding or misusing the API. Instead, we could pass the current JSON path instead of the node, like other API's do (`onEditable` and the `onChange` of templates). Maybe also rename `selectedElements` to `selectedPaths` in the code, and to make it easier to reason about, always pass an array with 1 or multiple paths?

3.  Nice that you've added some validation in `_processContextMenuPlugin`!

4. Currently you've implemented `toolbarPlugins` for the tree mode only. I think we also need it for other modes. We will need a way to configure in whether a toolbar button must be displayed in a specific mode or not. Could be by introducing an extra config option `modes: [...]` which defaults to all modes, or by passing the current mode with to the callback function that creates the button, so it can decide to return `null` when in a mode where the button is not applicable.

5. Can you add an example on how to use toolbar plugins, context menu plugins, and multi context menu plugins in the examples section? (Basically move `test_plugins.html` to examples, rename it and cleanup a bit where needed).

6. Currently all actions show green "+" icons (see screenshot). Instead, we should display no icon by default. And probably implement a way to specify custom icons yourself.
image

7. Currently, when the text of a menu doesn't fit, it's aligned on the next line which looks quite weird (see "Submenu+click" in the screenshot above). We should do some CSS fixes there.

8. Multi-select nodes by dragging is currently broken in the editor, so I can't test whether the plugins work there. I haven't looked into the issue itself yet.

dhughes-xumak:
> Currently all actions show green "+" icons (see screenshot). Instead, we should display no icon by default. And probably implement a way to specify custom icons yourself.

Commit bde68cf adds a `jsoneditor-plugin` class and the corresponding styles to remove the default '+' icon.

Currently, it sets the icon background to transparent for context menus and #d3d3d3 for toolbars.  If a plugin icon is added to the icon sheet, that should be used as the default.

Updated test_plugins to test and demonstrate icon overrides for plugins.

josdejong:
@dhughes-xumak That looks way better already :) 

I've just fixed the issue with the broken multiselection (see https://github.com/josdejong/jsoneditor/commit/2110d251bb9d950d7f4ff303b0f354f5b38ea58f, will do a release today), and have done some adjustments in the css of the context menu so it wraps long text neatly.

This resolves points 6, 7, and 8.

So the left over action points are now:

1. Rename click callbacks to `onClick` for all three menus.
2. Pass JSON path instead of Node to `onClick`
3. Implement support for toolbar plugins in all modes
4. Add an example in the examples section
5. Integrate undo/redo functionality into the plugins

josdejong:
(the merge conflict that's there now is just two new blocks of CSS inserted at the same location)

josdejong:
@dhughes-xumak this PR is still open, are you still interested in completing it?

dhughes-xumak:
No, sorry.  It would be great to see this included in the project, but I will not be completing it.

josdejong:
Ok clear! At least we have a nice start :)

If anyone is interested in finishing this feature please let me know.

PR #445, opened 2017-08-01
Enable displaying a node schema type instead of its value

ofrebourg:
Added option 'valueDisplayMode' to choose what the node value of a tree should indicate. If set to 'value' (default) it displays the node value, as it used to do. If set to 'schema' and a schema has been defined in the options, it displays the schema type of the node. If set to 'schema-if-null', it only displays the schema type if the value of the node is null.

### DIFF
--- a/examples/14_json_value_display_mode.html
+++ b/examples/14_json_value_display_mode.html
@@ -1,0 +1,132 @@
+<!DOCTYPE HTML>
+<html>
+<head>
+  <title>JSONEditor | JSON schema validation</title>
+
+  <link href="../dist/jsoneditor.css" rel="stylesheet" type="text/css">
+  <script src="../dist/jsoneditor.js"></script>
+
+  <style type="text/css">
+    body {
+      width: 600px;
+      font: 11pt sans-serif;
+    }
+    .jsoneditor {
+      width: 100%;
+      height: 250px;
+    }
+
+    table tr th {
+      text-align: left;
+    }
+  </style>
+</head>
+<body>
+<h1>Value display mode</h1>
+<p>
+  This example demonstrates the different value display modes:
+  <table>
+    <tr>
+      <th>value</th>
+      <td>displays the value of a node</td>
+    </tr>
+    <tr>
+      <th>schema</th>
+      <td>displays the schema type of a node</td>
+    </tr>
+    <tr>
+      <th>schema-if-null</th>
+      <td>displays the schema type of a node if its value is null</td>
+    </tr>
+  </table>
+</p>
+
+<h2>Value</h2>
+<div class="jsoneditor" id="jsoneditor1"></div>
+<h2>Schema</h2>
+<div class="jsoneditor" id="jsoneditor2"></div>
+<h2>Schema if null</h2>
+<div class="jsoneditor" id="jsoneditor3"></div>
+
+<script>
+  var schema = {
+    "title": "Example Schema",
+    "type": "object",
+    "properties": {
+      "firstName": {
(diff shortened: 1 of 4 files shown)


josdejong:
Thanks Olivier, I will look into your PR within a few days.

josdejong:
It's an interesting idea to give the user information about the expected type of value (based on the JSON schema). Thanks for adding an example too which demonstrates the feature. 

I was just thinking a bit more about you solution. It may not be needed to implement this as an option `valueDisplayMode`, plus a method and a callback on change. If the only purpose is to inform the user, it may be enough to replace the text of the placeholder ('value') with the JSON schema type (like 'number').

I'm not sure about the usecase for "valueDisplayMode: 'Schema'", where you can't edit the values, can you elaborate a bit more about the use cases you have in mind?

ofrebourg:
Interesting idea to replace the text of the placeholder with the JSON schema type. I hadn't thought of that.
I had in mind the idea of adding a switcher (like you have for the mode) to switch between the different valueDisplayModes. But if a switcher isn't necessary, your proposed solution might be enough.
Now, if you wanted to replace the text of the placeholder with the type, you'd currently get something like
image
instead of
image
The style can be changed, but as you can see, with the placeholder solution you don't get any validation of the type. Not sure whether this would be important for people or not.

I created the 'schema' option after the 'schema-if-null' because it was easy to do and thought some people might want it. This could come in handy if you had the switcher I mentioned above. Otherwise - without the switcher - you could use 'schema-if-null' and set all values to null.

ofrebourg:
I have another feature in mind. I know it's not part of this pull request but it's kinda related so I thought I would get your thoughts on that.
At the moment, if you have a schema containing more fields than the json object has, the missing fields do not appear in the tree view. I believe you get a validation error on the root but the missing field itself does not appear. If we had the option of showing the field type, as in this PR, it would be a great feature (at least for the project I'm working on) to be able to display all missing fields.
Diffing/merging the json and the schema could be outside of jsoneditor, but it would be a useful feature to have in jsoneditor.

josdejong:
I can't come up with use cases for the switcher, and if you don't have concrete use cases either I propose not to implement it until use cases for this pops up. I think though that in practice you don't want to _switch_ between value or schema, but you want to _enrich_ the value with information from the schema. A placeholder text is the simplest one. We could also look into a popover which gives all JSON schema details.

The solution with the placeholder has indeed a limitation in that you _only_ see the placeholder when the field is empty, not when the field contains `null`. I think that would be ok. A popover would not have this limitation. What do you think?

About the missing JSON schema fields: can you please open a separate issue to discuss that idea?

ofrebourg:
Agree, the switcher doesn't seem very useful after all.
Regarding the placeholder solution, my problem with it only being visible when the field is empty is that I don't know any way of initialising a jsoneditor view with an empty field. If I set a field to null, it appears as null. If I set it to undefined, the field simply doesn't show. For the project I'm working on, emptying a value manually (by clicking each value and emptying it) wouldn't be acceptable. I haven't tried emptying them programmatically. Is there any way to initialise a jsoneditor view with an empty field at the moment?
A pop over can be useful in some cases, but for one of my use cases, I'm looking for a jsoneditor read-only tree/view that displays a schema at all times without requiring hovering fields.

I'll open a separate issue about the JSON schema fields.

josdejong:
> If I set a field to null, it appears as null. If I set it to undefined, the field simply doesn't show. 

Yes that's a bit weird, there is indeed no "real" way to communicate an empty field. You can set it as an empty string for example.

Issue #447, opened 2017-08-17
Switching modes does not work in Shadow DOM


tomalec:
If `JSONEditor` is created in Shadow DOM (v1), switching modes does not work. The event is blocked by https://github.com/josdejong/jsoneditor/blob/c51db94f8863b421cf599c510fd553640f4bf68a/src/js/ContextMenu.js#L243-L251

as it attaches event listener to the `window` and from this perspective `event.target` is not longer a context menu item, but it's enclosing shadow host.

Live preview here http://jsbin.com/gucimed/3/edit?html,js,output (check in browser with Shadow DOM v1, like Chrome)
//cc @Mihaiii

PR #448, opened 2017-08-17
Add listeners to list (shadow) root instead of window,

tomalec:
make context menus work in Shadow DOM.

Fixes https://github.com/josdejong/jsoneditor/issues/447

### DIFF
--- a/src/js/ContextMenu.js
+++ b/src/js/ContextMenu.js
@@ -3,6 +3,14 @@
 var util = require('./util');
 
 /**
+ * Node.getRootNode shim
+ * @param  {Node} node node to check
+ * @return {Node}      node's rootNode or `window` if there is ShadowDOM is not supported.
+ */
+function getRootNode(node){
+    return node.getRootNode && node.getRootNode() || window;
+}
+/**
  * A context menu
  * @param {Object[]} items    Array containing the menu structure
  *                            TODO: describe structure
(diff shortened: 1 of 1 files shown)


josdejong:
Thanks for this fix @tomalec ! 

I think there is one more occurrence of `window`, in `ContextMenu.prototype.hide` where the event listeners are cleaned up again. I think that should be adjusted too.

tomalec:
Thanks for the review.
Is it better now?

josdejong:
Looks good, thanks for updating!

josdejong:
Your fix is now available in `v5.9.4`

tomalec:
Great! Thanks a lot!

josdejong:
@tomalec it turns out that your solution didn't work out exactly as we expected, and the global listeners where never cleaned up again causing the editor not responding to mouse clicks anymore after using the context menu once. I've fixed this in `v5.9.5` (see https://github.com/josdejong/jsoneditor/commit/d5ba54890265ca5d89e3d85404f92f5df1ab5619), can you please check whether the fix still works for shadow DOM too?

tomalec:
Thanks it works fine! Sorry for missing that out :|

BTW, Just to note, there is no 5.9.5 tag yet at https://github.com/josdejong/jsoneditor/releases

josdejong:
:thumbsup: I had missed it to. 

I've also created the `v5.9.5`, I had completely forgotten that.

PR #459, opened 2017-09-07
Display a dropdown for enums inside composite schemas

hachichaud:
Hi!

When trying to display a dropdown for an enum inside a composite schema (oneOf/allOf/anyOf), the dropdown would not be displayed.

Fiddle with the issue:
https://jsfiddle.net/gyhw9s2q/1/

Fiddle with fix:
https://jsfiddle.net/gyhw9s2q/

I fixed it by looping through the multiple schemas.
Note that this will return the last matching childSchema between all composites.


josdejong:
Thanks @hachichaud, looks like a neat fix

PR #463, opened 2017-09-13
Fix FindSchema for nested composites

hachichaud:
Hi again,

Turns out my previous fix wasn't enough to handle cases of nested composites :/
Fiddle with the bug:
https://jsfiddle.net/59h2cobx/2/

Fiddle with the fix:
https://jsfiddle.net/59h2cobx/3/

It will still output the last schema that matches: see https://jsfiddle.net/59h2cobx/4/ and https://jsfiddle.net/59h2cobx/5/

Cheers!

### DIFF
--- a/src/js/Node.js
+++ b/src/js/Node.js
@@ -2056,13 +2056,13 @@ Node._findSchema = function (schema, path) {
       if (typeof key === 'string' && childSchema.properties) {
         childSchema = childSchema.properties[key] || null;
         if (childSchema) {
-          foundSchema = childSchema;
+          foundSchema = Node._findSchema(childSchema, path.slice(i, path.length));
         }
       }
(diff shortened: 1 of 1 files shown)


josdejong:
Thanks! 

I expect to do a release coming weekend.

hachichaud:
Nice, thanks for the heads up :)

josdejong:
Just FYI: it turned out after your fix for JSON Schema, the JSONEditor didn't work anymore without defining a schema... I've fixed that in this commit: https://github.com/josdejong/jsoneditor/commit/de71499269c1d2ae3b54ae547427729211ffa5f7#diff-0d1bf82adf85971cf3b7e338e96fe882L2001

josdejong:
Your fix is now available in `v5.9.6`, thanks again!

PR #486, opened 2017-11-20
 Display a dropdown for enums inside pattern properties

alquist:
Fixes non-displaying dropdowns in enums when nested inside patternProperties:
```                 "actions": {
                    "type": "object",
                    "patternProperties": {
                        "^.*$": {
                            "type": "array",
                            "items": {
                                "type": "object",
                                "properties": {
                                    "xpath": { "type": "string" },
                                    "type": { "enum": ["string", "integer", "boolean"] },
                                    "value": { "type": "string" },
                                    "delay": { "type": "integer" },
                                    "mode": { "enum": [ "get", "set" ]},
                                    "target": { "type": "string" },
                                    "event": { "enum": ["click", "redirect", "fail"] }
                                },
                                "required": ["xpath"]
                            }
                        }
                    },
                    "additionalProperties": false
                }```


josdejong:
Thanks Ondrej, nice fix!

I will change the base of the PR to `develop`

Btw there is no need to commit changes in `./dist`. I only update that for a release, though I will soon remove it completely from the repo (it was only needed for bower which is now deprecated).

PR #490, opened 2017-12-02
Include caseSensitive option for autocomplete.

israelito3000:
Include caseSensitive option to search for autocomplete options.
Documentation and last example was updated.


josdejong:
Thanks Israel, looks good and works good!

One thought: right now, when you are typing entries with "wrong" casing, the entries in the dropdown are changed to match the wrongly cased input. I think it may make more sense to leave them as they really are: display them with their real casing. What do you think?

israelito3000:
You are right, I fixed in the last commit.

josdejong:
Thanks Israel, works like a charm :+1:

PR #492, opened 2017-12-21
Add caret position to text mode status bar

meirotstein:
+ enable status bar in text mode


josdejong:
Thanks Meir, looks good!

- There is no need perse to push the changes in `dist/jsoneditor.js`, I normally only update the files under `dist` when doing a release. (We'll drop the `dist` folder from the github repo soon anyway, it was only needed for Bower which is deprecated now). No need to adjust your PR for this though, it's no big deal.
- I think the use of `this` in `updateDisplay()` and the way you bind to `this` to the function can maybe be simplified a bit. How about something not using `this`? Pass regular arguments to the function (make it a pure function), or define and use `var curserInfoElements = this.curserInfoElements` instead of `var me = this` and `.apply`? Or create `updateDisplay` as a regular prototype method and bind it in the `_updateCursorInfoDisplay` method.

josdejong:
I just merged your #495, now there is a merge conflict in the `dist/jsoneditor.js`, that illustrates what I tried to explain with not including updated versions of files in `dist` in PR's LOL

meirotstein:
thanks for your review!

- I simplified my implementation so the `bind` in not needed - I simply used the `me` mapping in `updateDisplay()`, the `me` is used anyway in the timeout callback.
- sorry for the `jsoneditor.js` - it slipped under my radar :)

josdejong:
Thanks for the update, looks good!

PR #499, opened 2018-01-08
In the View mode, open links in the same tab by default

LEW21:
They'll still get opened in a new tab when the user presses Ctrl (like in the other modes).

### DIFF
--- a/src/js/Node.js
+++ b/src/js/Node.js
@@ -2151,7 +2151,6 @@ Node.prototype._createDomValue = function () {
       // create a link in case of read-only editor and value containing an url
       domValue = document.createElement('a');
       domValue.href = this.value;
-      domValue.target = '_blank';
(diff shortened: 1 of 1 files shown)


josdejong:
Hm, yeah. This is inconsistent with the regular behavior of clicking links (click opens in the same tab ctrl+click opens in a new tab).

There is only on way to open a link in the tab in Tree Mode. Not sure if that's an issue though. At least I think your fix will make the current behavior more consistent with default browers behavior, thanks!

LEW21:
Thanks!

It'd be nice to find a way to open links in the same tab in the Tree mode too, but I don't know how to do it in a user-friendly way. Also, when editing, accidentally opening the link in the same tab may lead to data loss, so maybe it's safer this way.

josdejong:
Agree

Issue #470, opened 2017-09-30
API for getting caret selection


josdejong:
@meirotstein has implemented a status bar where you can see the caret location. A discussion popped up there (https://github.com/josdejong/jsoneditor/issues/443#issuecomment-331626325) to create an API to get the current selection.

I think this API will contain a method `getSelection` and a callback `onSelectionChange`, and it probably needs a different implementation for the modes `tree`/`form`/`view` vs `text`/`code`.

meirotstein:
IMHO - I think that for `text` and `code` the callback should return the selected text and cursor positions (row/col for start and end)

As for `tree`, `form` and `view` - here the selected node should be returned, so the callback 
signature is different.

Perhaps it should be separated into different functions to avoid confusions? also I'm not sure that a `get` function is the most recommended here as it usually means a single value and not a registration of a listener, what about the following suggestions:

```
//text and code
onTextSelectionChanged(function(text, startPos, endPos){
...
})

//tree, form and view
onNodeSelectionChanged(function(node){
...
})
```

josdejong:
> Perhaps it should be separated into different functions to avoid confusions?

Yes I think so.

I like your proposal for `onTextSelectionChanged` and `onNodeSelectionChanged`. Instead of passing `node` to `onNodeSelectionChanged` though I think we can pass the *path* of the current node. It is also possible to have a multi-selection, so maybe the syntax should have a start and an end, something like:
```js
//tree, form and view
onNodeSelectionChanged(function(startPath: string[], endPath: string[]) {
...
})
```
Maybe there is also need to know exactly where the cursor is in the node: `field` or `value`, and maybe even the cursor position in the editable div, then we need to pass more information, something like:
```js
//tree, form and view
onNodeSelectionChanged(function(startNode: Object, endNode: Object) {
...
})
// where startNode and endNode have a structure:
// {
//   path: string[],
//   selection: {
//     input: 'field' | 'value'
//     start: number,  // start character index 
//     end: number     // end character index
//   }
// } 
```
Ideally the API should be a bit similar to for example `onEditable` which also passes a "node" object.

YuriyChaban:
josdejong, did you implement functionality for getting caret selection in API?
I need help with getting caret position and when click on some button in place where caret located, some json adding in that location in CODE editor

LIke this example:

http://i.prntscr.com/mKblu6WESCSUh1Vu4iCcng.png

if you have any idea how can i do this, please write me. Thank you.

[...]

PR #504, opened 2018-01-25
API for caret selection

meirotstein:
The PR adds the following new capabilities:

- `onTextSelectionChange` - new `option` for adding a callback for text selection change in `text` and `code` modes.
expected callback function signature is:
```
/**
  * @param {String} text selected text
  * @param {{row:Number, column:Number}} startPos selection start position
  * @param {{row:Number, column:Number}} endPos selected end position
  */
  function onTextSelectionChange(text, startPos, endPos) {
    ...
  }
```
- `onNodeSelectionChange` - new `option` for adding a callback for Nodes selection in `tree` mode.
expected callback function signature is:
```
/**
  * @param {Array<Node>} nodes selected nodes
  */
  function onNodeSelectionChange(nodes) {
    ...
  }
```
- `getTextSelection()` - returns the current selected text and selection range for `text` and `code` modes.
- `setTextSelection(startPos, endPos)` - set text selection for `text` and `code` modes.
- `getNodeSelection()` - returns the current selected Nodes for `tree` mode.
-  `setNodeSelection(node1, node2)` - set nodes selection for `tree` mode.

Additions are documented in details in `api.md`

@josdejong please note that the API for `onNodeSelectionChange` and `getNodeSelection` is a bit different from what we discussed because it returns a list of the selected nodes and not only the start and the end nodes, it makes more sense to me because probably it will be more useful to the utility consumer.


meirotstein:
#470

josdejong:
wow, Meir, that looks like a huge step! I will review your code soon.

josdejong:
First, sorry for my late reply Meir, I was working hard on a large next release of `mathjs`.
 
I really like your implementation, nice job! Some feedback points:

1. I think in the end you can move the file `test_selection_api.html` to the examples folder. Possibly extend it with demoing `setNodeSelection`.

2. I was thinking again about the naming for these functions. The method names containing "*node*" in it may not be meaningful for people. Maybe we should name them `onSelectionChange`, `getSelection`, and `setSelection`. That would be consistent with other methods `get`, `getText`, `set`, `setText`.

3. In the documentation for `setNodeSelection`, should the parameters be named `startNode` and `endNode` instead of `node1` and `node2` for a bit more clarity?

4. I agree that the user in the end probably wants to have the complete list with selected nodes rather than the start and end. But that's just one of two use cases: the other is that the user wants to get the selection simply to persist it and restore on browser reload, or sync the selection state with other browsers or with other parts of the same application. I think that we can have both: 
    - Let `getNodeSelection` and `onNodeSelectionChange` pass the start and end node, so they will be consistent with `setNodeSelection` where you have to pass start and end. (which is also consistent with the equivalents for text/code mode).
    - Provide a util function to generate the list with selected nodes from the start/end selection. This could look something like:

      ```js
      onNodeSelectionChange(text, startNode, endNode) {
        var selectedNodes = editor.nodesFromSelection(startNode, endNode);
        // ...
      }
      ```
      And maybe a version expanding into all selected child nodes is useful too?
      ```js
      onNodeSelectionChange(text, startNode, endNode) {
        var recurse = true;
        var selectedNodesAndChilds = editor.nodesFromSelection(startNode, endNode, recurse);
      }
      ```
5. Right now, the functions for getting/setting node selection pass the internal Node objects around. These nodes aren't really nice to be exposed to the user. It may invite users to (mis)use undocumented, the internal API, and it's not serializable (which we need for example when we want to persist the selected state, or we want to sync it between browsers). The API so far does not expose these nodes, but instead something serializable like the path. I think it will be nice if the contents of the node selection (`startNode` and `endNode`) would be an object containing a path, like 
`nodeStart={path:['array', 2]}`. Similar to the API of `onEditable`. Such an API with simple paths will make it easier to _do_ something with it (like manipulating some JSON in the internal data structure of an application according to the current selection). What do you think?

meirotstein:
@josdejong - thanks for the feedback! - I will change my code accordingly.

Some comments:
**Point # 2**: I agree that  `onSelectionChange` etc is more clean API, however it will make the callback signature complicated as it have to cover both text and tree modes, so the start and end parameters will be in different structure for each mode and it should to be investigated in order to understand the current mode (or pass the mode as parameter and so on..), so maybe it better to think of a better name - maybe `onTreeSelectionChange` ?

**Point # 3**: The reason for the `node1` and `node2` names is that actually this API not force the order of the nodes, you can pass the end node on the first parameter and the start node in the second and it will still select the range.

**Point # 4**: got it. however as the `nodesFromSelection` is a utility method perhaps it should be available on the `JSONEditor` class and not on the instance? (i.e. static). 
e.g.:
```js
var selectedNodes = JSONEditor.nodesFromSelection(startNode, endNode);
```

**Point # 5**: very good point, I also was not comfortable with the fact that all the logic is based on node instance and cannot be serialized.

josdejong:
**Point 2** sorry I did not mean a single function for both text and tree, but `onSelectionChange` for tree mode (similar to `get`, `set`), and  `onTextSelectionChange` for text mode (similar to `getText`, `setText`).

**Point 3** ah, ok. Also with `startNode` and `endNode` you can swich order :). in most text editors, the end is the place where the caret is, and can be at the start or the end of the selected text. If you think it's better to name them `startNode`, `endNode` I'm ok with that though I still think `startNode` and `endNode` are more clear from a user point of view.

**Point # 4** We can put it as a static util function, but you will need to pass the full JSON object too in order to be able to generate `selectedNodes`, so you get something like
```
JSONEditor.nodesFromSelection(editor.get(), startNode, endNode)
```
I think it will be easier to use when creating it as a method on an editor instance, so you don't have to pass the JSON object:
```
editor.nodesFromSelection(startNode, endNode)
```
 
**Point # 5** :+1:  curious to see your adjustments

meirotstein:
Got it. will adjust my code soon.

meirotstein:
@josdejong sorry for the delay here, having a quite busy days. I hope to make a progress soon.

A small comment in regard to point #2 - 
`treemode.js` already expose an API for selection under the 'setSelection' name, although this is a non-public API and quite technical one, renaming it might be considered as a breaking change..
What do u think?

josdejong:
> @josdejong sorry for the delay here, having a quite busy days. I hope to make a progress soon.

No worries, I have that all the time :D 

> A small comment in regard to point #2 -
> treemode.js already expose an API for selection under the 'setSelection' name, although this is a non-public API and quite technical one, renaming it might be considered as a breaking change..
> What do u think?

Thanks for being careful here! I expect the current internal `setSelection` not to be used externally, it requires passing an object with some DOM fields and a Range, it's not easy to use. It may not be too much work to keep the editor backward compatible: rename the current function to `setDomSelection` or something like that, and on top of the new `setSelection`, test whether the first argument is an object with properties `dom` and `range` (simple duck typing), and if so throw out a `console.warn('deprecated bla bla')` and redirect to `setDomSelection`.

meirotstein:
@josdejong I have added all the changes we discussed - sorry that it took me so long (took the advantage of some flights I had recently to work on it :) )

josdejong:
That's good news Meir, thanks! I will review your latest changes soon, I expect this week.

josdejong:
Impressive Meir! I really like your work, I think we're almost there.

Some feedback points:

1. Ordering of `start` and `end`. right now, the `start` is always the top side of the selection, and `end` always the bottom side. I think start and end should correspond with the way the user created the selection: `start` should correspond with the node where the user started dragging, and `end` where the user stopped dragging. So when dragging upwards, the `start` of the selection should be at the bottom. What do you think?
2. In `tree` mode, when selecting just one line, there is no `end` in the selection. I think `end` should always be returned and simply contain the same path as `start` (same to how selection in text mode works)
3. I really like the returned selections from the `getSelection` and `getTextSelection`, with an easy to handle `path` and the `value`. I'm not sure about the added value of `field`, since you could also retrieve that by simply picking the last item from the path. What is your idea with this property `field`?
4. The function `getNodesByRange` is nice, I like that it's decoupled now from the selection functions. Right now it returns internal Nodes of the editor though (which no part of the public API does so far, and will raise a lot of questions if we expose it). I think it may be more consistent and usable to return an array with SerializableNodes instead?

meirotstein:
Thanks for the feedback! as for your points:

1. We have to separate between the code and tree editors, in code/text nodes this is technically very hard to know the direction of the selection, as this information is not passed by the ace editor nor by the selection api. 
In tree mode it is possible to do that, as a developer I'm thinking how would I prefer to get the information.. wondering if the selection direction can be somehow useful - I'm not really sure. anyhow I can do that.

2. Makes sense. I will change it.

3. Well, this information indeed duplicated in some way. I had in mind that the serializable node should represents a json node, and a json node has field (or key) and value, so it will be more convenient to use it that way.. anyhow it is not a problem - if you think different I will remove it.

4. I understand your point in regard to the public API, anyhow I think that the main benefit of this utility is that it returns the real nodes, because as a developer I cant think of any usage in getting the serialized node list.
having the real node in hand can be much useful as it allows to expand, collapse, scrollTo etc.
Maybe the right direction would be to go over the current Node and define what can be added to the external API, once it will be decided we can think of also adding `getNodesByRange` to the current API (of course as part of different PR :) ) - at the moment I suggest not to add it at all.

josdejong:
1. Ai, that's a pity if Ace editor does not support that. I will have a look at the API of Ace editor too, see if I can some way to figure out the real start/end. Maybe there is a way to compare the start/end against the current position of the caret or some other trick to figure it out.
2. :+1:
3. That makes sense indeed. I think it's important to keep the API and data structures as small and simple as possible though, so I indeed prefer leaving out `field` since it's available in the `path` too. One other thought: including the `value` itself could be a performance issue when working with a large document and a large selection, but let's keep it like this for now, we can always extend the selection methods with an option to include/exclude `value`.
4. I totally agree that there is a lot of useful stuff in the Nodes that you could hook into, but it also opens up a lot of possible issues/misuse since the internal API isn't very neat. I agree not to expose the Node right now, and it's a good idea to address opening up (part) of the internal nodes in a separate PR. The use case that I had in mind is that `getNodesByRange` makes it easy to get all selected contents for example to copy it to somewhere, since the `onSelectionChange` event only gives you the path and value of the start and end, not all nodes in between.

josdejong:
@meirotstein I've done some testing with Ace editor, and I think we can use `getSelection` instead of `getSelectionRange`: this returns an object with amongst others property `anchor` and `lead` (containing `row` and `column`) which I think is what we're looking for.

josdejong:
I see there are even methods `getSelectionAnchor` and `getSelectionLead` :D

meirotstein:
@josdejong - I have added some changes according to your comments. pls review.
Thanks!

josdejong:
Thanks for the updates, I hope/expect to review your latest changes coming Wednesday.

josdejong:
@meirotstein I've reviewed the latest version of your work, and tested it thoroughly in the `15_selection_api.html`. It looks and works really solid, thanks. I like the current API a lot, it's simple to use and understand.

Two small last things that crossed my mind:

1. Maybe we should change the signature `onTextSelectionChange (text, start, end)` to `onTextSelectionChange(start, end, text)` (having `text` as last argument instead of first) so it's more consistent with `onSelectionChange(start, end)`?

2. What should `JSONEditor.getSelection()` return if there is no selection? Right now it returns `{}`, but we could also return `null` or `{start: null, end: null}`. Do you have thoughts in that regard?

meirotstein:
@josdejong no problem, I will handle these two issues.

josdejong:
Awesome. Thanks again! Really happy with the result.

josdejong:
@meirotstein I've just released v5.15.0 containing the new selection API :tada:

meirotstein:
Cool!
Happy to contribute to this great project!

PR #509, opened 2018-02-09
pass empty string when no previous field value

43081j:
`textDiff` takes strings, so we should be ensuring `previousField` is empty string rather than `undefined`.

`previousField` can become undefined if you drag drop some items into new positions (for example, an array item to an object; it has no previous key).


josdejong:
Thanks for the fix James. I tried to reproduce the issue you describe, but no success. Can you give an example?

It's a good idea anyway to make the code robust in this regard, so I will merge your PR now. For future PR's, can you please create a PR to `develop` instead of `master`, and only commit the changes in `src` (not the `dist` folder)? See: https://github.com/josdejong/jsoneditor/blob/master/CONTRIBUTING.md

43081j:
btw FYI i got this error by moving an array item out of an array and into an object, the `previousField` was then undefined. i suppose it shouldn't be undefined in the first place, so if i get chance i'll have a look at what caused that

josdejong:
yeah I tried to reproduce it like you describe but no luck so far

PR #510, opened 2018-02-09
cache parent errors

43081j:
There's a rather bad memory issue in the treemode functions.

Essentially, it is possible to have multiple errors on one node and multiple errors down a tree of nodes. This means we can end up in the situation where we append an error for the node's parents multiple times.

In cases where we have a large schema, a large amount of errors, we end up appending a huge amount of parent errors.

To combat this, i've gone ahead and added a cache which we check before adding an "invalid parent" error.

### DIFF
--- a/src/js/treemode.js
+++ b/src/js/treemode.js
@@ -435,6 +435,8 @@ treemode.validate = function () {
     }
   }
 
+  var parentErrors = [];
+
   // display the error in the nodes with a problem
   this.errorNodes = duplicateErrors
       .concat(schemaErrors)
@@ -443,7 +445,11 @@ treemode.validate = function () {
         // original entries last
         return entry.node
             .findParents()
+            .filter(function (parent) {
+              return parentErrors.indexOf(parent) === -1;
+            })
             .map(function (parent) {
+              parentErrors.push(parent);
               return {
                 node: parent,
                 child: entry.node,
--- a/src/js/treemode.js
+++ b/src/js/treemode.js
@@ -436,23 +436,28 @@ treemode.validate = function () {
   }
 
   var errorNodes = duplicateErrors.concat(schemaErrors);
-  var parents = errorNodes
+  var parentPairs = errorNodes
       .reduce(function (all, entry) {
           return entry.node
               .findParents()
               .filter(function (parent) {
-                  return all.indexOf(parent) === -1;
+                  return !all.some(function (pair) {
+                    return pair[0] === parent;
+                  });
+              })
+              .map(function (parent) {
+                  return [parent, entry.node];
               })
               .concat(all);
       }, []);
 
-  this.errorNodes = parents
-      .map(function (parent) {
+  this.errorNodes = parentPairs
+      .map(function (pair) {
           return {
-            node: parent,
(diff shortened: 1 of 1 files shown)


josdejong:
Thanks James! I think your PR solves `// TODO: dedupe the parent nodes` on line 459 in the `validate` function, right? If so, this TO DO can be cleaned up :)

https://github.com/josdejong/jsoneditor/blob/master/src/js/treemode.js#L459

Is it an idea to rewrite your solution into a small `uniq` function similar to the one from lodash, to make the code cleaner? (right now the code is a bit of a mix between a functional and declarative approach).

I've changed the base of your PR to `develop`. There are now some merge conflicts since I just merged your other PR and both have changes in dist. Can you remove the changes in dist from your PR?

43081j:
hmm we can't just `uniq` the resulting errors because we want to avoid creating those errors in the first place.

if we generate 100s of parent errors then filter them down to only a couple, we still had to allocate 100s initially. we could filter the parents out based on their node/message existing in the array already but its a bit unclean looking still

josdejong:
It's indeed less efficient to first generate errors and then filter most of them away again. I think it though that for "hundreds" of errors this will cost only a couple of microseconds and is neglectable. I always try to balance easy to understand code against good performing code. I don't expect a bottleneck here so optimizing right now at the expense of readability of the code may be premature optimization.

It's not a big deal though, let's not overthink this. If you want to keep your current solution I can live with it. What do you think?

43081j:
while usually that would be true, it can actually be a huge problem when `allErrors` is true in AJV.

if you have a recursive schema or something with a large `anyOf`, you can quite easily end up with thousands of errors produced by AJV. this already makes it easy to reach the maximum heap size in node and chrome, but they recommend we just turn `allErrors` off or use their async calls when that happens.

anyhow, it means if it does produce so many errors, we will essentially duplicate that amount here before filtering down to the ones which matter.

ill see if i can clean it up slightly though and comment here again soon

josdejong:
ok  :+1:  in that case it sounds like it makes sense to smartly optimize.

maybe rewrite to an plain old `for` loop then so there is no weird mix between functional and declarative programming?

43081j:
@josdejong i just pushed a refactor of that block. would you mind seeing if it works for you & looks any better than what it was? im on the move at the min so haven't tried it myself yet

josdejong:
Thanks @43081j  for the update and sorry for my late reply.

This is a neat approach! Your solution *almost* works, though there is a small issue with `child: entry.node,` not being defined (so you cannot click on the parent error to open and go to the related child). Can you have a look into that?

43081j:
you make a good point i totally overlooked 🙈 

because this de-dupes parent errors, it means we can only have one referenced child of potentially many children. should we just choose the first one?

43081j:
@josdejong see if this works for you. i had to use tuple-like arrays basically (`[parent, node]`) so we keep hold of the context node.

means parent errors will always point to the first erroring child i think

josdejong:
Thanks for the update James, neat solution!

josdejong:
Your fix is now available in `5.14.0`

PR #512, opened 2018-02-12
allow null autocomplete results

43081j:
The docs state:

> Can return null when there are no autocomplete options.

so we should be able to:

```js
getOptions(text, path, input, editor) {
  return null;
}
```

or

```js
getOptions(text, path, input, editor) {
  return Promise.resolve(null);
}
```

However, we can't currently do this because the promise-check will throw an exception: `typeof result.then === 'function'` (`result` is null).

### DIFF
--- a/src/js/treemode.js
+++ b/src/js/treemode.js
@@ -1216,13 +1216,18 @@ treemode._onKeyDown = function (event) {
           setTimeout(function (hnode, element) {
               if (element.innerText.length > 0) {
                   var result = this.options.autocomplete.getOptions(element.innerText, hnode.getPath(), jsonElementType, hnode.editor);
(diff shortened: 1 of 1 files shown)


43081j:
p.s. ill rebase these PRs to develop if you want (just saw your other comment)

josdejong:
Thanks, nice fix! Would be great indeed if you can change the base to `develop` and remove the files in `dist` from the PR.

43081j:
here you go. i wasn't sure if you wanted the dist files building. makes life easier to not have to do those

josdejong:
Thanks :+1: . Yeah the dist files are just an unfortunate left-over from from the bower era ...

PR #514, opened 2018-02-20
Basic translate

mariohmol:
Hi there.. awesome module.. thank you!

I'm using this and needed to have in my language, so i did this changes.. if you think this can be useful.. feel free to merge!

bests

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,9 +67,14 @@ with bower:
 
 #### More
 
-There is a directive available for using JSONEditor in Angular.js:
+
+There is a directive available for using JSONEditor in AngularJS:
 
 [https://github.com/isonet/angular-jsoneditor](https://github.com/isonet/angular-jsoneditor)
+
+Directive for Angular 5.x as well:
+
+[https://github.com/mariohmol/ang-jsoneditor](https://github.com/mariohmol/ang-jsoneditor)
 
 
 ## Use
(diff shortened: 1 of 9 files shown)


josdejong:
wow, this is awesome @mariohmol , looks good!

I think to make this complete, we could implement an option `lang` (which defaults to `navigator.language`), and implement a way to pass additional translations into JSONEditor without having to hack the source code, for example via a new option `locales: {...}`. Would you be interested to implement these options?

mariohmol:
Nowdays its getting the default lang by navigator language, but the translations are indeed in the source code. I will make this paremeter now to include more locales as a parameter

josdejong:
Cool, thanks! 

I think using `navigator.language` as default is a great solution, but there should be an option to force usage of a specific language too (ie. an option `lang`). Some applications couple a language to a user profile rather than a machine. And some sites may only support English and want to force that in all used components in order to have a consistent site. What do you think?

mariohmol:
I included in this PR, 2 options:

*  `languages` to insert or override translations keys 
* `language` property to force a language.

mariohmol:
Did a small improvement... guess now is everything good to go

mariohmol:
Added the language support on angular component 

https://github.com/mariohmol/ang-jsoneditor/commit/68877f0c3b7ae6648d89b4c57521fffe786f5de8

josdejong:
Looks great, thanks for updating the docs and adding an example!

After merging I will refactor your code to ES5, there isn't yet a transpiler in the project setup to convert to ES5.

josdejong:
@mariohmol your translation solution is now available in `v5.14.0`, thanks again.

mariohmol:
Thanks! I upgraded the ang component to this version and did the demo with buttons where can change language to test.

https://raw.githubusercontent.com/mariohmol/ang-jsoneditor/master/src/assets/printDemo.png

PR #515, opened 2018-02-28
jsoneditor-icons.svg is explicitly in a relative path to be compliant…

landru29:
jsoneditor-icons.svg is explicitly in a relative path to be compliant with webpack plugin "file-loader"

file-loader is, actually, interpreting `img/jsoneditor-icons.svg` as absolute path. To ensure the path to be relative, just put `./`


josdejong:
Thanks Cyrile, that's a nice and logic improvement!

PR #536, opened 2018-06-05
Allow to add custom elements to tree node

trsh:
Basic usage:

```
  var options = {
    onDomTreeElement: function(dom, tr){
      var td = document.createElement('td');
      var btn = document.createElement('button');

      btn.value = "Press me";
      btn.onclick = function(){
        console.log(dom.value);
      }

      td.appendChild(btn);      
      tr.appendChild(td);
    }
  };
  var editor = new JSONEditor(container, options);
```

### DIFF
--- a/src/js/JSONEditor.js
+++ b/src/js/JSONEditor.js
@@ -90,7 +90,7 @@ function JSONEditor (container, options, json) {
         'ajv', 'schema', 'schemaRefs','templates',
         'ace', 'theme','autocomplete',
         'onChange', 'onEditable', 'onError', 'onModeChange', 'onSelectionChange', 'onTextSelectionChange',
-        'onAddCustom', 'escapeUnicode', 'history', 'search', 'mode', 'modes', 'name', 'indentation', 
+        'onDomTreeElement', 'escapeUnicode', 'history', 'search', 'mode', 'modes', 'name', 'indentation', 
         'sortObjectKeys', 'navigationBar', 'statusBar', 'languages', 'language'
       ];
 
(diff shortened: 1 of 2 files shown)


josdejong:
Thanks @trsh, I will review your PR within a couple of days

trsh:
@josdejong tnx

josdejong:
Thanks @trsh , I think allowing to extend the rendered DOM tree can be really powerful.

So far, JSONEditor doesn't expose anything from the internal nodes or DOM to the public API (see `onEditable`, `onSelectionChange`, `getOptions` in  `autocomplete`). So this would be the first that exposes the internal DOM, which may open up a lot of questions and requires documentation of the inner workings. This can be a good choice, but maybe we can explore whether we can come up with an API that's as powerful as your proposal but doesn't expose internals? I think it should roughly allow:

- be triggered on creation of a node
- be triggered on the update of a node
- pass the current path and value of the node
- should return DOM to be inserted 
- should specify where to insert the custom DOM element: before or after any of the elements drag button, action menu button, expand button, field, value.

What do you think?

trsh:
@josdejong I think you know best. Im new to this. 

I need to add my custom element to row badly, so thats why I started this. 

Im not sure I can implement all those things as you wrote, as my schedule is very tight for next month or so, and i'm familiar with the architecture of this lib. :/

josdejong:
Yes that's true. I try to keep discussions about one topic in one issue, that works quite well in most cases. And also, it's very easy to reference other issues on github. But sometimes you have to search in multiple issues which is not ideal.

Issue #550, opened 2018-07-02
Feature: extend event behaviour


cristinabarrantes:
Hi Jos,

   I would like to extend the functionality with a new API option to extend behaviour when an event is triggered. The reason to do it is because I would like to use internal structures and code when some type of events are triggered. I have done the changes locally so if you agree, I open a PR.

Best regards,
Cristina

josdejong:
That could be interesting Cristina. Can you explain a bit more about the actual events that you needed and implemented?

cristinabarrantes:
Hi Jos,

I think that the clearest way to show you this, is with a PR so I have just opened [one](https://github.com/josdejong/jsoneditor/pull/551).

Tell me if you want more information about it.

Thank you,
Cristina

josdejong:
This issue should have been closed already long time ago...

cristinabarrantes:
Ok, thank you!

PR #551, opened 2018-07-03
Add API functionality to extend event behaviour.

cristinabarrantes:
Let the possibility of extending events behaviour (among the ones managed by jsoneditor).
Function is provided as an option and will be triggered after node event management. This is added at the end of a node event execution, not replaced. It means that actions for that event will be executed and after that, if this function is provided, will be executed.
For example, if we want to extend the behaviour when a click event is received providing information about the element clicked and the path, we can provide this function and use internal structures and internal code to do it. An example is provided with the PR (examples/16_extend_event_api.html).


josdejong:
Thanks for your PR @cristinabarrantes ! 

I'm a little behind but hope to checkout your PR soon

cristinabarrantes:
No problem @josdejong. Thank you for the reply!

josdejong:
Sorry for the delays @cristinabarrantes ...

Before putting more effort in your PR, let's first get a clear plan on if/how/why we want this functionality.

Here some feedback:

1.  I understand how what you mean in #550. Still, can you explain your use case, what do you do with the events? I think such a method can make sense, but before accepting a new feature, I want to have the use case really clear.

2.  Thanks for working out the docs in detail, including an example!

3.  It looks like `this.extendEvent = ` is unused, is this redundant code?

4.  Instead of :

    ```js
    node.onEvent(event, this.options.extendEvent);
    ```

    I think you can also do:

    ```
    node.onEvent(event, this.options.extendEvent);
    if (this.options.extendEvent) {
      this.options.extendEvent(node, event);
    }
    ```

    That way the `node.onEvent` method stays simpler and decoupled.

5.  I think we need to think a bit more about the name of this function, maybe `onEvent` makes more sense?

6.  This would be the first method exposing the internal `Node` objects. This internal `Node` object is totally undocumented and would require quite some documentation about what you can and can't do with it. If possible, I would like to avoid exposing this internal class in the public API, and instead, pass an object like 

    ```
    {field: string, value: string, path: string[]}
    ```

    As is done by `onEditable` and something similar in `onSelectionChange`.

    Whether that makes sense depends on the use cases we foresee (see 1.)

cristinabarrantes:
Hi @josdejong ,

   I reply to your points:

1. Use case: in the future I think that I will need more functionality but nowadays, I only need to know the path when a user clicks on a field or value. So, although I was thinking in the future when I implemented this, I have changed my mind to think in present.

2. You are wellcome :-)

3. Sorry, removed

4. My approach was to have a single method where you have centralized all the actions that you are going to take for an event but I agree with you in decoupling. Now I have changed the code after your comments and I am not sure if you prefer the onEvent call in treemode or Node (there are pros and contras). I have included it in Node.js but I'll wait for your opinion about it. This is the code:

```javascript
  if (typeof this.editor.options.onEvent === 'function') {
    this._onEvent(event);
  }
```

5. Perfect. onEvent.

6. I was looking at onSelectionChange but it didn't work for me. I have looked at onEditable after your comment and it works so I have changed the code and updated the PR.

Thank you for your comments! Waiting to hear from you again :-)

Apart from that, doing tests I have seen that field for arrays is undefined due to this code in Node.js:

```javascript
  if (this.type == 'array') {
    // array
    this.childs = [];
    for (var i = 0, iMax = value.length; i < iMax; i++) {
      childValue = value[i];
      if (childValue !== undefined && !(childValue instanceof Function)) {
        // ignore undefined and functions
        child = new Node(this.editor, {
          value: childValue
        });
        visible = i < this.MAX_VISIBLE_CHILDS;
        this.appendChild(child, visible, notUpdateDom);
      }
    }
    this.value = '';
  }
```

Is there a reason for this?

Best regards,
Cristina

josdejong:
Thanks for the update in the PR, looks really good now!

1. and 6. if your use case is to get the current path when a user clicks on a field or value, I would expect that you could use the existing `onSelectionChange` event? Why can't you use it, what do you miss exactly? And what future usage scenarios do you see for yourself?

2., 3., 4., and 5. :+1:

7. I'm not entirey sure what you mean about the undefined `field` in case of Arrays. It is undefined when the root of the JSON object is an array, it is only defined for nodes which are the child of an object. Does that explain what you see?

josdejong:
ping @cristinabarrantes

cristinabarrantes:
Hi @josdejong ,

   Sorry for the delay, I have been out some days.

   Replying to your questions:

1. I would expect that you could use the existing onSelectionChange event? Why can't you use it, what do you miss exactly?

There are several reasons:
1.1 onSelectionChange does not work for view mode. I don't know if there is a bug or if documentation is incorrect (based on api.md, only tree is allowed but based on JSONEditor.js: tree, view and form)
1.2 I require the event, onSelectionChange is only restricted to selection change event and I want any event (for example, click)

2. And what future usage scenarios do you see for yourself?
My future scenarios (I have not seen if it is possible to do it without any change) are:
2.1 For view mode, change style for an element (field or value) and cursor when focus
2.2 Show a contextual menu when you press right click on a node

3. I'm not entirey sure what you mean about the undefined field in case of Arrays. It is undefined when the root of the JSON object is an array, it is only defined for nodes which are the child of an object. Does that explain what you see?

I have seen that it is not only when the root of the JSON object is an array. That is the behaviour for any element in an array, not only root. For example, for next json:

```javascript
  json = {
    "myarray": [1, 2, [3, 4, 5]]
  };
```

If I click on 1:
- path: <myarray,0> (using getName() in getPath() in Node.js)
- field: <undefined> (using getField() in Node.js)

This is not an issue, just a comment.

cristinabarrantes:
Hi again @josdejong ,

   Apart from my previous comment, I have realized that path provided is not enough because I need to know the exact path (including if a node is an array) and with current path, that cannot be identified. For example, you get same:
- path: 0
for next jsons clicking in first element:
- json1: ["zero"]
- json2: {"0": "zero"}

So path does not distinguish 0 as an object name or 0 as an array index.

First of all I thought about including the array index between brackets in path but it is not enough because, this is the counterexample:
{"[0]": "zero"}

So I have included an extended path in which type is informed for each element. I have just updated PR.

Best regards,
Cristina

josdejong:
Currently, the paths in other functions like `onEditable` return an Array containing the path, like `['myArray', 5, 'name']`. You could indirectly know whether something is a property or an Array entry by checking whether it is a string or a number. But you could also simply check by getting the real JSON object and (recursively) see whether every parth of the path is an Array or an Object. Would that suffice?

josdejong:
Thanks for sharing more about your use case and future use cases. 

Case 1.1 Really makes sense. It makes sense to me to add support for selection when in view mode. 
Case 1.2: Yes, but *why* do you need other events such as click, what are you going to do with it?

The future you mention are interesting too. There may be alternative solutions though and I think we should address them on their own as soon as it is an actual issue. 

What I sense in general in your comments is that you simply want to be able to hack into the editor via events, and adjust the HTML DOM, styling, and menus as you like. However, the JSONEditor code internally is quite a mess and far from flexible enough to allow easy hacking into the editor. (hence my effort to implement a `v6` from scratch). I'm a bit afraid that if we open up the internals of the editor for hacking into it, that it will cause a lot of trouble because of internal quirks. That's why I'm a bit reluctant in this regard, and try to search for other solutions which give you more flexibility without opening up the internals.

cristinabarrantes:
Hi @josdejong ,

Regarding your first comment, I agree. I can use getPath and check if typeof is a number. I have changed it and updated the PR.

Regarding your second comment, sorry if the impression is that I want to hack it, I would like to extend the functionality to be usefull for more people with your help. That is the reason why I have not done my own fork and I try to add it to jsoneditor project (and of course it is better for new functionalities and bug fixing if using jsoneditor project). I would analyze if enabling of the selection API in view mode is enough for me and I'll ltell you.

Thank you,
Cristina

cristinabarrantes:
Hi @josdejong ,

   I have been analyzing selection and I think it does not cover my requirements. I will explain them to you because perhaps I am missing something:
- Nowadays, I would like to show the path when clicking a node in view (just clicking, not selecting). That is what you can see in example 16.
- In the near future, I would like to show a contextual menu when clicking a node (field or value)

   In any case, what I would like is to extend the behaviour when clicking a node. Is it ok current PR for you?

Regards,
Cristina

cristinabarrantes:
Hi @josdejong ,

   Have you seen my replies? Do you need further explanation?

Thanks,
Cristina

josdejong:
Thanks for your reminder and sorry for my late reply. I've left all open pr's and issues of my open source projects as is for a couple of days to work focused on a great new feature of the JSONEditor, allowing to update existing JSON without losing state. 

I will checkout your latest feedback and changes soon.

cristinabarrantes:
Ok, thank you for your response.

Great new feature!

Best regards,
Cristina

cristinabarrantes:
Hi @josdejong ,

Regarding your questions:

1. I think that it is redundant because type would be the same as event.type, wouldn't it?
2. I did it because I thought that it was clearer for the user, don't you think so?

Regards,
Cristina

josdejong:
1 and 2: ok makes sense, let's keep it as it is right now then. If needed we can expand the functionality in the future. 

I will merge your work now, thanks for all your effort and patience @cristinabarrantes !

josdejong:
Your `onEvent` callback is now available in `v5.22.0` :tada:

cristinabarrantes:
Thank you very much for your help!!!

PR #579, opened 2018-09-15
Don't package CONTRIBUTING.md or examples folder

MarkHerhold:
If you don't want this, please just close.

Thanks for the awesome module!

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,7 +1,6 @@
 bower.json
 CONTRIBUTING.md
 downloads
-examples/
 misc
 node_modules
 test
--- a/.npmignore
+++ b/.npmignore
@@ -1,5 +1,7 @@
 bower.json
+CONTRIBUTING.md
 downloads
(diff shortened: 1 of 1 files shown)


josdejong:
Thanks Mark, I think CONTRIBUTING is indeed not needed in the package. Examples on the other hand are very relevant I think.

MarkHerhold:
I always look at it from the other perspective of a package being downloaded tens of thousands of times by CI servers for deployments. Less is sometimes more. At any rate, I'll make the change.

josdejong:
Yes it's good to try make packages as small as possible. For me optimizing for the user comes first, and machines come second. That means having docs and examples at your disposal everywhere :).

I prefer leaving the `examples` in, but `CONTRIBUTING.md` can be removed, that has no value in the context of a published package.

josdejong:
Thanks for the update Mark.

PR #584, opened 2018-10-03
Fix #582: parse error annotations not always up to date in code editor

### LINKED ISSUE #582: changeAnnotation event problem
It is possibly and fairly easy to replicate this bug

open 03_switch_mode.html
switch to code mode
replace everything with "{},"
now remove the comma (with backspace for example) and close after restore the comma (by pressing comma or ctrl+z)
now if you timed that correctly you will have a mismatch between annotations and the status bar (parseErrorIndication)

I think this is a bug, I haven't been deep diving into the code to see whats happening but I suspect the debouncer to be part of the problem? Or is there any asynchronous race condition?

If you vary the timing between removing the comma and restoring it you will notice you get different results, listening on changeAnnotation event I can see the following scenarios happening

1: as above, I get two events which seems out of order, the first event containing the error and the second one with an empty list, the time between the events are less than 50ms, ~20-30ms

2: if you restore the comma extremely fast you will also get two events, the same as above, but in the right order, same timing between them as scenario 1, ~20-30ms

3: if you remove it and wait for about 150-200ms before putting it back you get 3 events, first empty, second containing error and third empty as well, first event coming at time 0, second ~50ms after and third >~150ms after the first one

4: if you wait any longer you get 4 events, starting with 2 empty ones for removing the comma and same as in scenario 3 second one comes ~50ms after the first one, when restoring the comma you get 2 more events, 50ms timing between them, first empty, second with errors as expected

I can also confirm that the mismatch does not seems to happen for brace in itself, https://thlorenz.com/brace/
you can test the exact same thing in the upper right box, never got it to misbehave

josdejong:
Fixes #582 

@meirotstein I think I may have found a solution for #582. I'm not sure though whether this may break other things, can you double check?

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,6 +2,11 @@
 
 https://github.com/josdejong/jsoneditor
 
+## not yet released, version 5.24.7
+
+- Fix #582: parse error annotations not always up to date in
+  code editor.
+
 
 ## 2018-09-12, version 5.24.6
 
--- a/src/js/textmode.js
+++ b/src/js/textmode.js
@@ -531,7 +531,10 @@ textmode._emitSelectionChange = function () {
 }
 
 textmode._refreshAnnotations = function () {
-  this.aceEditor && this.aceEditor.getSession().setAnnotations();  
+  var session = this.aceEditor && this.aceEditor.getSession()
+  if (session) {
+    session.setAnnotations(session.getAnnotations())
+  }
 }
 
 /**


meirotstein:
The refresh function target is to redraw the updated validation errors - this fix is redrawing the existing ones.
anyhow I was not able to reproduce the original problem reported on #582 - can you please give some hint how you did that? 10x!

josdejong:
> can you please give some hint how you did that? 10x!

Yes to reproduce I open a code editor with the following contents

```
{}
```

Then test correct working by pressing <kbd>,</kbd> at the end of the contents (changing the contents to `{},`), and waiting a second. Now you see an error annotation left from the contents.

To test the wrong behavior, start with the contents `{},` and press <kbd>backspace</kbd> and then <kbd>,</kbd> quickly after each other, changing the contents to `{}` for a fraction of a second and directly after changing it back to `{},`. Now there is no error annotation left from the contents (though there is an error message bottom right).

What I noticed:
- If you disable debouncing of the validation function, everything works correct
- `setAnnotations` is invoked twice: 
  - in the correct case it's first called like `setAnnotations()` and after that as `setAnnotations([{...}])`
  - in the wrong case it's first called like `setAnnotations([{...}])` and after that as `setAnnotations()`.This has to do with the debounced validate triggering the call `setAnnotations()` (without arguments) after the other trigger.

meirotstein:
@josdejong I would not recommend disabling the debouncing... as you said - it is there for a reason.
I have done some debugging and it seems that your approach is the safest - however you only have to re-set the annotations of `error` type, in the previous solution my concern was that in some scenarios the validation annotations will not refresh as expected - see my PR to this branch in #587 .

**A short explanation:** 
ace uses its own json mode to handle json errors - it done async using web workers therefore by using debounce we got into this edge case.
The validation annotations however are handled in our code.
The're no situation where we'll have error and validation annotations at the same time, therefore I overriden the `ace.session.setAnnotation` implementation to handle both scenarios.

the `_refreshAnnotations` is a private function that removed the error annotations - the idea behind it was that `setAnnotations` will be retriggered by ace in case of a parse error. 
In this edge case the debouce 'swalloed' some of these triggers so the error wasnt there as expected. this fix adds additional safty layer by adding the existing error annotations on refresh, so it that case it will keep showing the error annotation, if exists.

josdejong:
Ah, thanks for the explanation. Your solution in #587 makes sense. Thanks for adding extra comments in the code explaining it.

PR #588, opened 2018-10-09
Adds ability to hide sort and transform options

### LINKED ISSUE #580: Hide the Sort and Transform buttons
In [version 5.19.0](https://github.com/josdejong/jsoneditor/blob/master/HISTORY.md#2018-07-11-version-5190), sort and transform options were added. While these are really cool options, I would like to hide them from users to keep the interface simple.

Are there plans to allow hiding the sort/transform buttons via the options object?

Thanks for your time and the awesome module!

tanmayrajani:
Closes #580

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -340,7 +340,6 @@ Constructs a new JSONEditor.
 
   The default language comes from the browser navigator, but you can force a specific language. So use here string as 'en' or 'pt-BR'. Built-in languages: `en`, `pt-BR`. Other translations can be specified via the option `languages`.
 
-
 - `{Object} languages`
 
   You can override existing translations or provide a new translation for a specific language. To do it provide an object at languages with language and the keys/values to be inserted. For example:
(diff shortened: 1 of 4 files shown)


josdejong:
Thanks @tanmayrajani , I will review your PR coming weekend.

josdejong:
Sorry for the delay @tanmayrajani, I had some busy weeks.

Your PR looks nice and simple, thanks! And happy to see you updated the docs accordingly :)

A few feedback points:

1. I think we should also disable the sort and transform buttons in the context menu (available when you click the action button left from an Object or Array), see method `Node.showContextMenu`.
2. I'm thinking about the naming of these two options (`hideSort` and `hideTransform`). I generally try to avoid negations in options. And I think we should try to be a bit consistent with similar (boolean) options like `history`, `search`, `navigationBar`, `statusBar`, and `colorPicker`. How about naming the new options `sort` and `transform` (most consistent with the existing options), or `enableSort` and `enableTransform` (maybe most clear)? What do you think?

tanmayrajani:
I went with `enableSort` and `enableTransform` but after the commit, I now feel like they're better off being consistent with other options. 

Also, I was a bit skeptical about "Ability to display the Sort option" wordings. Any other suggestions there?

josdejong:
:+1: thanks @tanmayrajani

pedro-j-morais-alb:
hi. someone knows how to hide Repair JSON button in Code mode? enableRepair = false isnt working for me. thx

josdejong:
There is no option to disable the repair button, but you can hide it using CSS.

PR #596, opened 2018-10-26
Create option to hide the blue main menu bar on top 

### LINKED ISSUE #581: Create option to hide the blue main menu bar on top
```
var editor = new JSONEditor(dom, {
                            mode: 'code',
                            navigationBar: false,
                            statusBar: false,
                        });

```
statusBar is hidden, but navigationBar isn't.

image

tanmayrajani:
Fixes #581 

- also contains some typo corrections, JSDoc corrections etc.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -234,6 +234,10 @@ Constructs a new JSONEditor.
      - Can return an object `{startFrom: number, options: string[]}`. Here `startFrom` determines the start character from where the existing text will be replaced. `startFrom` is `0` by default, replacing the whole text.
      - Can return a `Promise` resolving one of the return types above to support asynchronously retrieving a list with options.
 
+- `{boolean} mainMenuBar`
+
+  Adds main menu bar - Contains format, sort, transform, search etc. functionality. True by default. Applicable in all types of `mode`.
+
 - `{boolean} navigationBar`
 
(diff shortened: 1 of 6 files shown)


josdejong:
Thanks @tanmayrajani, your PR looks good! Thanks for all the little code corrections too :)

Two remarks:

1. In textmode.js, I think it would be neat to move the code to create the `var poweredBy = ...` inside the first `if (this.options.mainMenuBar) {...}` block so there is only one such block instead of two.

2. When disabling the `mainMenuBar`, the contents of the editor do not fill up the whole container: at the bottom, there is space left with the height of the main menu. There is some CSS involved here (top padding and margin). This is from before the time we have `display: flex` ;) Can you look into that?

tanmayrajani:
Thanks for the review, @josdejong! I'll work on these changes soon. Having a bit busy week :)

tanmayrajani:
Sorry for the delay, I've pushed my changes in above commit. Let me know if I missed something, it's a big changes so it's quite possible. I've tested the CSS changes against all modes, with and without `mainMenuBar`, `statusBar` and `navigationBar`.

josdejong:
Thanks Tanmay, I hope to test your latest changes tomorrow.

josdejong:
I've tested `mainMenuBar` extensively, with different modes and combinations of `navigationBar` and `statusBar`, and all works like a charm, thanks again @tanmayrajani !

josdejong:
The new option `mainMenuBar` is now available in `jsoneditor@5.26.0`

PR #600, opened 2018-11-12
Fix deepEqual for null values

DullReferenceException:
When doing a `.update` on an editor that was showing an object and replacing it with `null`, a runtime exception occurs. This is because the `deepEqual` check doesn't account for the fact that `typeof null === 'object'`, so it tries to access properties of a null value.

### DIFF
--- a/src/js/Node.js
+++ b/src/js/Node.js
@@ -144,7 +144,7 @@ Node.prototype.getName = function () {
 
 /**
  * Find child node by serializable path
- * @param {Array<String>} path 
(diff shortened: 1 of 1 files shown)


josdejong:
Thanks for the fix @DullReferenceException, this is one of these nice pitfalls in JavaScript :D

josdejong:
Should be fixed now in `v5.26.1`

DullReferenceException:
Thanks for the quick turnaround, @josdejong!

PR #615, opened 2018-12-18
Add options to hide drag and menu buttons

antfu:
```js
const options = {
  dragButton: false,
  menuButton: false,
}
```

image

In some use cases, you may want to restrict user from editing the structure of the JSON object.


josdejong:
@antfu thanks. Yes it can make sense to restrict editing the structure. The mode `'form'` was created for this purpose, it removes these buttons and makes the fields readonly, so only values are editable. Doesn't the mode `'form'` address your concern already?

antfu:
I didn't notice there is the `form` mode when I was making this PR. The `form` mode works great! Thanks.

josdejong:
:+1:

PR #616, opened 2018-12-18
 i18n for ModeSwitcher

antfu:
The **ModeSwitch**'s text was hard coded. This PR fixed it.

https://github.com/josdejong/jsoneditor/blob/72ab51002c27381474522d4090de5ed0ad3c237f/src/js/ModeSwitcher.js#L17-L46

*Other change:*
**Unquoted** i18n keys and fixed multiple lines' indent. For better readability.

Before:
image

After:
image

### DIFF
--- a/src/js/ModeSwitcher.js
+++ b/src/js/ModeSwitcher.js
@@ -1,6 +1,7 @@
 'use strict';
 
 var ContextMenu = require('./ContextMenu');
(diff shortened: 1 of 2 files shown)


josdejong:
Thanks for this fix @antfu :+1: 

looks good to me. I like the conditional operator better with `?` and `:` on the start of the line but will leave it for now.

PR #620, opened 2018-12-31
Hide show error table

meirotstein:
As discussed in #606 (and on another issue that I can't find :smiley: ) - this PR allows toggling the visibility of the validation error tables for text and code mode
default behavior remains the same: visible for text mode, hidden for code mode.

### DIFF
--- a/src/css/statusbar.css
+++ b/src/css/statusbar.css
@@ -29,6 +29,7 @@ div.jsoneditor-statusbar > .jsoneditor-validation-error-icon {
   padding: 0;
   margin-top: 1px;
   background: url("img/jsoneditor-icons.svg") -168px -48px;
+  cursor: pointer;
 }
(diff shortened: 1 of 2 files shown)


josdejong:
Really neat @meirotstein, I like it!

Two small feedback points:

1. right now you have to click on the yellow icon to toggle visibility, I think it would be nice if toggling also works when clicking on the number showing the total number of issues?
2. when I try to run `npm run build` the minification fails for some reason. Not sure if it's my setup or something in your code changes that confuses uglifyjs.

meirotstein:
@josdejong - happy new year & thanks for your feedback!

1. done + additional fix to preserve table state after fixing parse error
2. fixed - it was an arrow function that mistakenly slipped in..

josdejong:
:+1:  thanks

we should some day put a transpiler into place so we can simply use arrow functions and other good stuff :)

PR #623, opened 2019-01-03
Update Node.js

tylerchen:
fix childSchema with $ref, and not display the select element on the child schema because of not found enum


josdejong:
@tylerchen is your PR #624 replacing this PR? Should we close one of the two?

PR #627, opened 2019-01-08
Allow user to set max node children in editor options

20goto10:
Adds a "maxChildren" option that allows the user to set the number of nodes visible in the tree before the "show more / show all" message appears. The default remains 100.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -3,6 +3,12 @@
 https://github.com/josdejong/jsoneditor
 
 
+## not yet published, version 5.27.1
+
+- Improved navigating deeply nested paths via the navigation bar, see #619.
+  Thanks @meirotstein.
+
+
 ## 2019-01-05, version 5.27.0
 
 - Implemented customizing object and array names via a new option
--- a/src/css/treepath.css
+++ b/src/css/treepath.css
@@ -2,11 +2,35 @@
 div.jsoneditor-treepath {
   padding: 0 5px;
   overflow: hidden;
+  white-space: nowrap;
+  outline: none;
+}
+
+div.jsoneditor-treepath.show-all {
+  word-wrap: break-word;
+  white-space: normal;
+  position: absolute;
+  background-color: #ebebeb;
+  z-index: 999;
+  box-shadow: 2px 2px 12px rgba(128, 128, 128, 0.3);
 }
 
 div.jsoneditor-treepath div.jsoneditor-contextmenu-root {
   position: absolute;
   left: 0;
+}
+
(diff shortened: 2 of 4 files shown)


josdejong:
Thanks @20goto10 :+1: Looks straight forward.

Two feedback points:

1. I think it would be better not to change `Node.prototype.MAX_VISIBLE_CHILDS`, which is defined globally and meant to be static. I haven't tested, but I expect the current solution to have an issue when you create two editors with a different option for `maxChildren`. Instead we can simply use the new `options.maxChildren` everywhere and when it's undefined, fallback on the  `Node.prototype.MAX_VISIBLE_CHILDS` (or even make it an internal constant instead of property). What do you think?
2. I'm not sure about the name for this new option. `maxChildren` sounds a bit like you cannot have more children. I like to keep option names short, but it's more important that they are clear. How about something like `defaultVisibleChilds`, `initiallyVisibleChilds`, `maxVisibleChilds`, or something like that?

20goto10:
Sure @josdejong , sounds good to me on both counts. I'll put in an update shortly.

20goto10:
I've put in an update for you. Note that I renamed all uses of "childs" to "children" (my apologies for the sudden expansion of the PR size, but it should reduce a bit of confusion). 

On the plus side I've tested this change with two JSONEditors in the same window, testing different and absent values of the maxVisibleChildren option, and everything behaved as it should (the second editor was not affected by the first, and fell back to defaults when the maxVisibleChildren option was unspecified).

josdejong:
Thanks for the update Ben.

1.   I understand the reason why you want to rename `childs` to `children`, consistency is important. I'm reluctant to make this change though, because even though this is all internal API, there may be people relying on it, which would cause their solution to break. So there needs to be a very good reason to rename this and I would prefer to make it a breaking change. I'm not sure this is important enough to make this change. We can also decide to name the new feature `maxVisibleChilds` so we have both consistency and no breaking changes. What do you think?
2.   I see you still adjust the globally shared `Node.prototype.maxVisibleChilds`, which very easily can cause issues (even though it may currently work correctly with multiple editors). How about introducing the following method on Node and use that where `maxVisibleChilds` is needed?
      ```js
      // private global constant
      var DEFAULT_MAX_VISIBLE_CHILDS = 100;
  
      Node.prototype.getMaxVisibleChilds() {
        return (this.editor && this.editor.options && this.editor.options.maxVisibleChilds)
            ? this.editor.options.maxVisibleChilds
            : DEFAULT_MAX_VISIBLE_CHILDS;
      }
      ```

20goto10:
Reopening this on a separate branch due to a merge issue I didn't have time to sort out...

Issue #619, opened 2018-12-30
Deeply nested path of properties not fitting in navigation bar


joshkmartinez:
When using *BIG* json files (mine is 2385 lines long) it is hard to see the list of nodes (under the search bar) when you select a node very deep into the json tree.  
like [**this**](https://i.imgur.com/1wWsZhC.png)
I think it would be helpful to make the list of nodes kind of like a carousel scroll thing so it is easier to see the list of nodes

josdejong:
Thanks @joshkmartinez , you're right, when your deeply nested in a document, the path may not fit in the navigation bar.

@meirotstein do you have some ideas to improve this? Wrap to the next line or show ellipsis halfway the path or something?

meirotstein:
@josdejong @joshkmartinez 
This is a bug. the navigation bar should always show the selected node - see [here](https://github.com/josdejong/jsoneditor/issues/443#issuecomment-332020010) the discussion we had about it in the past.
Let me look into it a bit, also if you think about a better approach - let me know.

meirotstein:
@josdejong I had a bit of miss understood here - I thought that the selected node is not visible - which is a bug. however, this is not the case here.

The current approach trims the beginning of the path if it is not fit the navigation bar space and currently there's no way to move along it.

What I have in mind is when the path is trimmed to add arrows that allow to 'scroll' along the path. I think that two arrows in each direction are required - one for scroll and one for 'jump to start/end', something like that:
![image](https://user-images.githubusercontent.com/2209844/50729139-58159f00-113d-11e9-92ea-738dd93df088.png)

What do you think?

josdejong:
@meirotstein sounds good, I'm ok with either solution you think is best. Curious what you come up with.

josdejong:
I forgot to close this issue: improved navigation path is available since some days now in the latest version thanks to the PR of Meir (#628).

PR #628, opened 2019-01-10
Add option for show all path on the navigation bar (tree mode)

meirotstein:
As per the discussion on #619 -
I have added the option to view the entire path of selection on the navigation bar.

This PR adds the following:
When there is no space to show the entire path, a button with an ellipsis will appear on the left side of the navigation bar:
image

Clicking this button will show the entire path on a floating area:
image

The floating area will be removed when the focus will be removed from it.


josdejong:
Thanks Meir! The screenshots look great, exactly what we're looking for. Unfortunately I can't get it to work when checking out your PR. I opened `07_json_schema_validation.html` and put some deeply nested JSON there, but the path is always left aligned and the overflowing part is simply hidden on the right.

In the code I see one issue (I guess unrelated): in the function `_onShowAllClick` there is a `var me = this` and `this` is used a couple of times. I think you meant to use the `me` which is already defined outside of the function, and use `me` instead of `this` inside the function?

meirotstein:
@josdejong that's odd... I have tested it again on my local env and it is working for me on Chrome, FF and Safari.. perhaps it is an IE issue?
Can you add a screenshot of what you see?

In regard to the `me` and `this` - indeed it's a bit confusing but this is not an issue, anyway I will make it more friendly

josdejong:
It could be some cache issue in my Chrome browser or something, will give it another try.

josdejong:
I guess I'm overlooking something but I can't get it to work.

I tried a slightly adjusted example 3, it looks like:

image

```html
<!DOCTYPE HTML>
<html>
<head>
  <title>JSONEditor | Switch mode</title>

  <!-- when using the mode "code", it's important to specify charset utf-8 -->
  <meta http-equiv="Content-Type" content="text/html;charset=utf-8">

  <link href="../dist/jsoneditor.css" rel="stylesheet" type="text/css">
  <script src="../dist/jsoneditor.js"></script>

  <style type="text/css">
    body {
      font: 10.5pt arial;
      color: #4d4d4d;
      line-height: 150%;
      width: 500px;
    }

    code {
      background-color: #f5f5f5;
    }

    #jsoneditor {
      width: 400px;
      height: 500px;
    }
  </style>
</head>
<body>

<p>
  Switch editor mode using the mode box.
  Note that the mode can be changed programmatically as well using the method
  <code>editor.setMode(mode)</code>, try it in the console of your browser.
</p>

<div id="jsoneditor"></div>

<script>

  var options = {
    mode: 'tree',
    modes: ['code', 'form', 'text', 'tree', 'view'], // allowed modes
    onError: function (err) {
      alert(err.toString());
    },
    onModeChange: function (newMode, oldMode) {
      console.log('Mode switched from', oldMode, 'to', newMode);
    }
  };

  var json = {
    "array": [1, 2, 3],
    "boolean": true,
    "null": null,
    "number": 123,
    "object": {"a": "b", "c": "d"},
    "string": "Hello World",
    "foo bar baz 1": {
      "foo bar baz 2": {
        "foo bar baz 3": {
          "foo bar baz 4": {

          }
        }
      }
    }
  };

  var editor = new JSONEditor(container, options, json);
</script>
</body>
</html>
```

josdejong:
(Tried both Firefox and Chrome, and checked that the code is from your branch and that I've created a build of that code by adding a console.log and checking that this pops up in the browser)

josdejong:
:joy: I really wasn't looking to break it, just to start it and enjoy this new feature.

I got it working when replacing removing all the spaces, in the example I posted above I changed the json document to:

```js
  var json = {
    "array": [1, 2, 3],
    "boolean": true,
    "null": null,
    "number": 123,
    "object": {"a": "b", "c": "d"},
    "string": "Hello World",
    "foobarbaz1": {
      "foobarbaz2": {
        "foobarbaz3": {
          "foobarbaz4": {
            "foobarbaz5": {

            }
          }
        }
      }
    }
  };
```

Can it have to do with the spaces?

meirotstein:
nope, I got it. silly bug :grimacing:  will submit a fix shortly

meirotstein:
@josdejong I just submitted additional fixes.
Indeed there was an issue with spaces as you assumed and found an additional issue - both were fixed.

PR #624, opened 2019-01-03
Update Node.js

tylerchen:
add both:
1. fix childSchema with $ref, and not display the select element on the child schema because of not found enum
2. add title from schema description to show the tips for user input

PR #633, opened 2019-01-16
fix childSchema can not access $ref and add title to display the schema description

tylerchen:
fix childSchema can not access $ref and add title to display the schema description


josdejong:
ow sorry I was confused with your different PR's, I've now merged #624, but it has the same code changes as this PR, right?

tylerchen:
Right, I am sorry about that because of the last commit create the new pr and I confused those prs.

josdejong:
No problem :+1:

josdejong:
Your fix is now available in `v5.27.1`

PR #640, opened 2019-01-31
i18n support zh_CN

long2ice:
i18n support zh_CN

### DIFF
--- a/src/js/i18n.js
+++ b/src/js/i18n.js
@@ -1,6 +1,6 @@
 'use strict';
 
-var _locales = ['en', 'pt-BR'];
+var _locales = ['en', 'pt-BR', 'zh_CN'];
 var _defs = {
     en: {
         array: 'Array',
--- a/src/js/i18n.js
+++ b/src/js/i18n.js
@@ -1,321 +1,321 @@
 'use strict';
 
-var _locales = ['en', 'pt-BR', 'zh_CN'];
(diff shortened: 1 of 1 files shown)


josdejong:
Thanks @long2ice! I will review and test your PR within a few days.

josdejong:
@long2ice I've tested your PR, looks good, thanks again! 

I have a few small remarks: 

1. Can you please change the indentation to 2 space? It' not consistent everywhere in the code base, but the intention is to use 2 space indenation everywhere.
2. Can you rename `zh_CN` to `zh-CN` to make it consisten with `pt-BR`?
3. It looks like `autoType` and `typeTitle` are not translated. This is also the case in the `pt-BR` translation. I see these fields defined the `i18n.js` file though. I think this is a bug that was already there in the editor. Would you like to try figure out why that doesn't work?
image

josdejong:
Thanks for the update, I will merge your work now.

@long2ice any interest in fixing the titles that don't work?

long2ice:
which test file?It looks like normal here.

long2ice:
@josdejong

josdejong:
That's odd, on my computer (both Chrome and Firefox) I still see the screenshot I posted [here](https://github.com/josdejong/jsoneditor/pull/640#issuecomment-460375774). Nevermind if you can't reproduce it, I will look into it.

PR #645, opened 2019-02-04
Set debounce interval

dverkhovsky13:
Sometimes user can not edit anything before onChange event is triggered, so I guess we need to add debounceInterval property that user can set.

### DIFF
--- a/src/js/JSONEditor.js
+++ b/src/js/JSONEditor.js
@@ -92,6 +92,8 @@ if (typeof Promise === 'undefined') {
  *                                                         in 'tree', 'view', or 'form' mode before 
  *                                                         the "show more/show all" buttons appear.  
  *                                                         100 by default.
+ *                               {Number} debounceInterval Debounce interval for keyboard input in milliseconds
+ *                                                         for onChange, onChangeJSON and onChangeText events.
  *
  * @param {Object | undefined} json JSON object
  */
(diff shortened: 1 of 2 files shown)


josdejong:
@dverkhovsky13 thanks for your PR. Can you explain your use-case where this is needed? Why can't the user edit anything during a debounce?

dverkhovsky13:
@josdejong Hi, the reason is that while I'm editting anything onChange event is triggering several times cuz of debounce with 150ms, sometimes user can type with lower speed or smth like that and after onChange event is triggered I need to put some data on a server and sometimes it's heppening 2 or 3 times before I finish editting.

josdejong:
Thanks for your explanation, makes sense to limit the amount of requests send back to the server. On the web site https://jsoneditoronline.org you have the same issue, there I've added a debounce of 1 second to store documents in local storage, and 4 seconds to store online. I think you could as easily debounce your onChange callback with the interval you want? 

The main reason that the debounce is there is to prevent onChange from firing with _every_ key press which can make the browser slow. It only acts on key presses and is not fool-proof, for example if you press Ctrl+D rapidly to delete a lot of items in an Array you will get an `onChange` event for every delete action also if they are within 150 ms.

What do you think? Would simply debouncing onChange work for you?

josdejong:
:+1: thanks for your PR anyway, will close it now.

PR #648, opened 2019-02-06
Allow enum with blank value as valid output by schema

pulges:
Adds an option to configure blank value for enum lists in shema.
Adds support for enum lists containing the blank value. (Currently double blank options where rendered in list)

```

var schema = {
  "title": "Example Schema",
  "type": "object",
  "properties": {
    "academicDegree": {
      "enum": [null, "B.S.", "M.S.", "Ph.D."]
    }
  }
};

var options = {
  mode: 'tree',
  schema: schema,
  enumDefaultValue: null
};

var json = {
  "academicDegree": null
};

var editor = new JSONEditor(container, options, json);
```

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -185,6 +185,12 @@ Constructs a new JSONEditor.
   Schemas that are referenced using the `$ref` property from the JSON schema that are set in the `schema` option,
   the object structure in the form of `{reference_key: schemaObject}`
 
+- `{String | null} enumDefaultValue`
+
+  Set default blank option value for `enum` lists in schema. `""` by default.
+  
+  If `enum` list contains the blank value, it will still be rendered as the frst option with title '--', but will be valid according to schema.
+
 - `{boolean} search`
 
   Enables a search box in the upper right corner of the JSONEditor. True by default. Only applicable when `mode` is 'tree', 'view', or 'form'.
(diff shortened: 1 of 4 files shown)


josdejong:
Thanks for your PR @pulges . 

Just thinking: shouldn't we simply remove the current behavior which always adds a blank value (and makes an empty string in that case)? So `"enum": [null, "B.S.", "M.S.", "Ph.D."]` will render a dropdown where the first one is displayed as "--" or blank, and `"enum": ["B.S.", "M.S.", "Ph.D."]` would render a dropdown with three options and no option to change a value to null?

pulges:
I thought about that, but the logic would lack then an option to display invalid/null value already present in json. Fron UX perspective it might work, if invalid option would be still rendered with '--', but with the validation not passing. If user picks another value '--' disappears from options. Thus user can not pick back invalid value.

josdejong:
That's a good point. Maybe when the current value is invalid, we could let the enum add an extra entry for this with a clear indication that this current value is invalid. Suppose the current value is "foo" and is invalid, the dropdown could render "foo (INVALID)" as first entry, and as you suggest, remove this entry as soon as the user picks a valid value.

johanbm1405-commits (approved):

johanbm1405-commits (approved):

PR #650, opened 2019-02-08
Add schema-based tooltip to field names

### LINKED ISSUE #321: Use schema titles as field names and descriptions as tooltips
Hi!

I've used your editor as a configuration editor in my web application. I needed the field text to be more descriptive, so I thought about using the "title" and "description" field of json schemas as field names and tooltips.

I've changed very few lines of code (diff attached below). The field itself is not changed, only the way it appears on screen. Sorting and searching also work. I've added two configuration options "showTitles" and "showDescr" to activate the behaviour.

The only significant change is the move of the _updateSchema() call from the updateDom() function to the appendChild() function. In the end I think that's a better place to put it anyway (unless I'm missing something).

Will you accept it as a patch?

Thank you for your great work

EDIT: Updated patch, had whitespace issue.
EDIT2: Updated patch again...changed blur() function.

[node_diff.txt](https://github.com/josdejong/jsoneditor/files/445291/node_diff.txt)

AdamVig:
Using the `title` and `description` properties from the schema, create
and set a tooltip on each field name. When the user hovers over a field
name, it will show the applicable information: title, description, both,
or neither, depending on what data is present in the schema.

Closes #321.

*Object with Title and Description*
image

### DIFF
--- a/src/js/Node.js
+++ b/src/js/Node.js
@@ -2466,8 +2466,6 @@ Node.prototype.updateDom = function (options) {
       domField.contentEditable = this.editable.field;
       domField.spellcheck = false;
       domField.className = 'jsoneditor-field';
-      // add title from schema description to show the tips for user input
-      domField.title = Node._findSchema(this.editor.options.schema || {}, this.editor.options.schemaRefs || {}, this.getPath())['description'] || '';
     }
     else {
       // parent is an array this is the root node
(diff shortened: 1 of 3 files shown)


josdejong:
Thanks @AdamVig , looks really neat!

I think this would replace setting of the title in `Node.updateDom`, right?

https://github.com/josdejong/jsoneditor/blob/4c6f2cdd4f467da5dbe0f38064143340b52577ae/src/js/Node.js#L2464-L2465

I guess it's best to set the title on a single place and `Node.updateDomField` is probably the most logical place.

AdamVig:
@josdejong My mistake, I didn't see that this feature had already been added in #624. Thanks for pointing it out. I will remove the redundant code.

josdejong:
Thanks again Adam! I've tested a bit more, all works smooth :)

PR #651, opened 2019-02-08
Fix Node._findSchema()

AdamVig:
The static method `Node._findSchema()` returns incorrect results sometimes. I added some tests that exercise the method to show the situations it fails in, and will hopefully be able to fix the method early next week.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,10 +15,40 @@
       "integrity": "sha1-ujpCMHCOGGcFBl5mur3Uw1z2ACg=",
       "dev": true
     },
+    "abab": {
+      "version": "2.0.0",
+      "resolved": "https://registry.npmjs.org/abab/-/abab-2.0.0.tgz",
+      "integrity": "sha512-sY5AXXVZv4Y1VACTtR11UJCPHHudgY5i26Qj5TypE6DKlIApbwb5uqhXcJ5UUGbvZNRh7EeIoW+LrJumBsKp7w==",
+      "dev": true
+    },
     "acorn": {
       "version": "3.3.0",
       "resolved": "https://registry.npmjs.org/acorn/-/acorn-3.3.0.tgz",
       "integrity": "sha1-ReN/s56No/JbruP/U2niu18iAXo=",
+      "dev": true
+    },
+    "acorn-globals": {
+      "version": "4.3.0",
+      "resolved": "https://registry.npmjs.org/acorn-globals/-/acorn-globals-4.3.0.tgz",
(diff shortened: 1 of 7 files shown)


AdamVig:
@josdejong I wouldn't merge this yet because the tests are failing. I will push further commits on this branch to fix the broken method.

josdejong:
Thanks @AdamVig for diving into this! I guess `jsdom` will be a devDepencency?

AdamVig:
@josdejong You are welcome. Yes, `jsdom` will be a dev dependency.

AdamVig:
@josdejong Ready for review.

josdejong:
Thanks, will do testing and review coming weekend. So I guess you found the issue making `_findSchema` not work correctly in all cases?

AdamVig:
@josdejong Thanks, sounds good.

Yeah, I found and fixed two issues:
1. With a property whose schema was in `patternProperties`, the function would always return the schema defined last in the `patternProperties` object. This was fixed by checking if the current property matches any of the regular expressions defined in the `patternProperties` object and returning the schema if the property matches.
2. With a schema that has both `properties` and `patternProperties`, a schema from within `patternProperties` would be returned sometimes for a property whose schema was in `properties`. This was fixed by adding a check to the `patternProperties` conditional.

josdejong:
wow, that's great!

AdamVig:
It is definitely possible that in fixing these two issues I may have created new issues, but since there are no pre-existing automated tests it is difficult to tell. I tested the library from this branch in an internal application at my company and it fixed the bug I was seeing, and seemed to work correctly otherwise, but I didn't conduct an in-depth test.

It might be better in the future to replace the `Node._findSchema` method with an external library that has its own suite of unit tests, because walking JSON schema accurately is difficult and prone to bugs. I found two options: [`json-schema-traverse`](https://github.com/epoberezkin/json-schema-traverse/) by the `ajv` guy and [`json-schema-walker`](https://github.com/cloudflare/json-schema-tools/tree/master/workspaces/json-schema-walker) by Cloudflare. Both are focused on traversal rather than finding subschemas, but it might be relatively easy to repurpose either of them to work for finding subschemas.

josdejong:
Thanks again starting to create unit tests, the library hardly has any unit tests, this helps a lot.

It would indeed be a good step see whether we can use an external library for all the JSON Schema related logic (like we do use Ajv instead of reinventing the wheel).

josdejong:
@AdamVig your fixes are now available in `v5.29.0`

Issue #652, opened 2019-02-08
turkish language support


beratpostalci:
I need Turkish language support for my application. #653

andreasseidler:
The turkish language support introduced duplicates in the "tr" json object, which breaks the usage of the jsoneditor in IE.

josdejong:
Support for Turkish is available in the latest version. The issue reported by @andreasseidler has been fixed too now.

PR #653, opened 2019-02-08
turkish language support is added - #652

beratpostalci:
I have translated the Turkish identities of all fields and have updated i18n.js file. #652


josdejong:
Thanks @beratpostalci !

PR #659, opened 2019-02-17
Add ability to add to/customize context menu

RobAley:
I see that the ability to add custom items to the context menu is in progress for version 6, however I'm guessing v6 is still some way away. 

This pull request introduces a simple way to allow customization of the context menu in the current version. It simply lets the user register an "onCreateMenu" function in options to which the contents of the menu "items" variable (plus reference to the current node) are passed before the menu is created from them.

I currently use it internally to add a custom menu item to display the current path as a JSON Pointer path (by pushing a new item onto items), and to alter the styling & text of some of the standard menu items depending on the current path (by altering existing item in items).

If this is something you want to incorporate, I can work up an example for the examples folder and additions to the api docs.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -146,11 +146,11 @@ Constructs a new JSONEditor.
   
   See also option `schema` for JSON schema validation.
   
-- `{function} onCreateMenu(items,path)`
+- `{function} onCreateMenu(items,{path})`
   
   Customize context menus in tree mode.
   
-  Sets a callback function to customize the context menu in tree mode. Each time the user clicks on the context menu button, an array of menu items is created. If this callback is set, the array is passed to this function along with the current path (if any). This function can customize any aspect of these menu items, including deleting them and/or adding new items. Each menu item is represented by an object, which may also contain a submenu array of items. See the source code of example 21 in the examples folder for more info on the format of the items and submenu objects.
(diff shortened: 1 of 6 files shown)


josdejong:
Thanks @RobAley! Yes makes sense not to await v6.

Some feedback:

1. Can you describe the new feature in the docs? Also, an example demonstrating how to use it would be great.
2. Right now you pass `node` as second argument, but so far the nodes are internal and undocumented API. Maybe we can instead pass an object containing the path of the current node, similar to `onEditable` and `onEvent`  and some other callbacks work. What do you think?

RobAley:
I agree with returning path rather than node, I can always find the node from the path if needed.

I've changed it to return the path rather than node. I've also added it to appendNodeFactory.js and treemode.js to capture (i think) everywhere a tree's context menu is created.

I've also added an example and put it in the docs.

Let me know what you think. Happy to change anything (or for you to do so).

josdejong:
wow, that's really nice @RobAley !! I love the example, it clearly demonstrates the power of customizing the menu, thanks.

Thanks for adjusting the API. I would love to see one more change there: change the second argument of `onCreateMenu` from being the path to being an object containing the path, like `onCreateMenu(items, node)` where node is an object `{ path: [...] }`. This will make it consistent with other functions like `onEditable`, and it makes it easy to expand the API in the future if needed by adding more properties to the `node` object.

RobAley:
Ok, I think I've changed that to how you mean. Do feel free to change it further if needed.

Thanks.

josdejong:
Thanks for the updates, I like this new feature :)

I've made some changes in the example to make it work on IE11 too.

josdejong:
`onCreateMenu` is now available in `v5.30.0` :tada:

RobAley:
Thats great, many thanks.

PR #664, opened 2019-03-04
Show examples from schema in field tooltips

AdamVig:
image


AdamVig:
@josdejong I'm not sure if you get notified about pull requests, so I am tagging you here as well. Let me know if this is redundant; I don't want to spam you. 😄

josdejong:
Thanks Adam, looks good on first sight, thanks for taking the time to even do all translations :)

I will review+test in more detail tomorrow or the day after.

I do get notified on every PR and issue. An extra message in a PR doesn't cause extra spam on my side, it's just one email thread in my mailbox, so no problem at all.

josdejong:
I've done some testing, all works like a charm. Thanks again @AdamVig!

josdejong:
Available now in `v5.31.0`

PR #665, opened 2019-03-05
Fixed using hyphens in the path of custom validation errors

tobiasfriden:
Using hyphens in the path of custom validation errors didn't work since the method `stringifyPath` previously added a dot to the start of anything which wasn't a number. This for example turned `fill-color` to `.fill-color` which is not a valid json path. Using bracket notation did not work either as `['fill-color']` was transformed to `.['fill-color']`. With this fix, using bracket notation works.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1706,8 +1706,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "aproba": {
           "version": "1.2.0",
@@ -1728,14 +1727,12 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "bundled": true,
           "dev": true,
-          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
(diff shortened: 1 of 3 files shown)


josdejong:
Thanks @tobiasfriden , good find!

Can you add a unit test for this so we prevent regressions in the future?

tobiasfriden:
@josdejong sure, will do.
However, I was wondering if you think it is necessary to check in `stringifyPath` if there is also a closing bracket in this case (and otherwise throw an error). In my case I can see it is done anyways in `parsePath` so depends on if `stringifyPath` is used for other purposes I guess.

josdejong:
I think the check for a matching closing parentheses `]` is not strictly necessary right now, but at some point in the future it could be that path that you parse has an external source and is not guaranteed to have a valid syntax. Since it's easy to add such a check it would be nice to add, but if you think it's overkill it's also fine with me to leave it out.

tobiasfriden:
@josdejong added the check as well as some unit tests for `stringifyPath`

josdejong:
Thanks for adding the unit tests @tobiasfriden.

One question: why would you pass `['foo', '["prop-with-hyphens"]']` as path? Wouldn't you instead pass `['foo', 'prop-with-hyphens']`and want the stringifier to turn it into `foo["prop-with-hyphens"]` (with either single or double quotes)?

tobiasfriden:
@josdejong that is indeed a good idea, my thought was to add myself the brackets to path element where they are required. But of course it would be better if the stringifier could do it automatically! I guess we only need to check if the path includes hyphens and in that case add brackets instead of the leading dot.
Are there any other special cases you know of that require the brackets? It might be good to include in the solution as well in that case.

josdejong:
:+1: I like it.

I think we have a similar issue when a path has spaces, and there easily be more cases I guess. Maybe we go defensive here: when the string contains only known, safe characters (regexp word characters `a-z`, `A-Z`, `0-9`, `_` and maybe `$`), then we use dot notation, and else we use the "safe" brackets+quotes notation?

tobiasfriden:
@josdejong that's a great idea. I'm not very experienced in writing regexp but I think this should be correct.

josdejong:
Thanks @tobiasfriden , looks good :+1:

josdejong:
This fix is now published in `v5.32.1`

PR #666, opened 2019-03-08
 Display schema defaults inline next to Nodes

AdamVig:
inline-defaults-v1


josdejong:
Thanks Adam, interesting idea to utilize JSON Schema defaults. Your movie explains it all :)

I'm thinking about how we can use defaults in a generic way. Most common way to display defaults is to show it as placeholder text when the field is empty. And what you often see is that non-default configuration have a distinct styling (like bold text, a different color, or a star icon on the right). I haven't seen your approach where you display the default value when the actual value is not the default, and I'm not sure if this is what everybody would like to see though I can understand that this is a useful approach. 

I think it is important that this feature is customizable via CSS, that you can change it or turn it off. And/or maybe we can extend `onClassName` to pass information about being default or not so you can adjust styling based on that. Or we could turn this in a callback function which allows you to generate some HTML to render right from the value when it's not the default. Just brainstorming here. What do you think?

src/js/Node.js
```diff
@@ -1908,6 +1910,33 @@ Node.prototype._getDomField = function(silent) {
...
+ * @private
+ * @returns {undefined}
+ */
+Node.prototype._updateDomDefault = function () {
+  // Short-circuit if schema is missing, has no default, or if Node has children
+  if (!this.schema || !this.schema.default || this._hasChilds()) {
+    return;
+  }
+
+  if (this.value === this.schema.default) {
```
kaelistus: Since you are setting all these things it would be nice if they are all ordered the same. Instead of title being set below text-context in the first if and at the top of the second if...

AdamVig:
@josdejong Thank you for the feedback! I will work on making it possible to disable the feature entirely and to customize the display of defaults.

As a side note, I don't see any documentation for `onClassName` other than [this example](https://github.com/josdejong/jsoneditor/blob/afead89e3b0638ca12be70fb953aa43be949bdd7/examples/20_custom_css_style_for_nodes.html).

The user experience for the feature is something I have definitely been struggling with. I agree with your feedback; my current implementation does not meet standard user expectations. I am going to map out the various states and try to come up with a new version of the feature.

Here are some resources about placeholders and default values in forms:
- <https://ux.stackexchange.com/questions/95655/are-placeholders-good-for-default-values>
- <https://ux.stackexchange.com/questions/75236/displaying-custom-vs-default-settings>

josdejong:
Whoops! Not sure how we forgot to document `onClassName`, I've added docs now via https://github.com/josdejong/jsoneditor/commit/1acd2d6116860e7f79709c44dde4e37f5fab8c1e.

Thanks for the pointers to these resources about defaults and placeholders, quite interesting! Placeholders may indeed not be the best approach. Trick is to have this information not get in your way but is available when you actually need it. Modern IDE's do this very nicely I think: when typing something, you see a popover displaying information about the method you're typing.

Just another idea: maybe we can show it in a popover, together with the JSON Schema description and examples. We could create a light gray question mark icon with a popover right from the value, similar to the screenshot below. And separately, add a class name when matching the default value, so you can easily customize styling for non-default values if you want.

image

AdamVig:
Hey @josdejong, thanks for the feedback. I ended up removing the styles for default values, leaving it up to the user to add them, and moving the display of default values into the existing tooltips, as you suggested.

Here is a new demo, with the following styles applied in my application (not the library itself):
```css
div.jsoneditor-value.jsoneditor-is-not-default {
	font-weight: bold;
	color: #008000;
}
```
inline-defaults-v2

Let me know what you think of this new version.

josdejong:
I love it!

I think it would be nice to update `07_json_schema_validation.html` a bit to show case this functionality.

Also it would be nice to have the ability to do custom styling for this somewhere in the documentation though I'm not sure what would be a suitable place for it.

AdamVig:
@josdejong I added more examples to the file you mentioned and added a documentation file about styling.

josdejong:
Looks good! Thanks again @AdamVig.

josdejong:
@AdamVig I've made some changes in example 07_json_schema_validation.html to demonstrate custom styling of default values. I noticed two small issues there, fixed that via 7c398aeef79789ab54d8d7468e87448b570819af. I think we also want to allow custom styling on a enum select box (like `gender` in the example), I haven't implemented that yet.

AdamVig:
Good catch! I'm glad you fixed that buggy conditional. The example looks good too.

josdejong:
I had a short try adding the classes to `select` box by simply checking in the `_updateDomDefault` method whether to pick `this.dom.select` or `this.dom.value`, but that doesn't work out of the box, but looking at the code it should work. Maybe something weird is going on in `_updateDomValue` or something. Will have a another look later (unless you beat me fixing this ;) )

josdejong:
This was what I was trying:

```js
Node.prototype._updateDomDefault = function () {
  // Short-circuit if schema is missing, has no default, or if Node has children
  if (!this.schema || this.schema.default === undefined || this._hasChilds()) {
    return;
  }

  const inputElement = this.dom.select
      ? this.dom.select
      : this.dom.value;

  if (!inputElement) {
    return;
  }
  
  if (this.value === this.schema.default) {
    inputElement.title = translate('default');
    util.addClassName(inputElement, 'jsoneditor-is-default');
    util.removeClassName(inputElement, 'jsoneditor-is-not-default');
  } else {
    inputElement.removeAttribute('title');
    util.removeClassName(inputElement, 'jsoneditor-is-default');
    util.addClassName(inputElement, 'jsoneditor-is-not-default');
  }
};
```

AdamVig:
Hey @josdejong, I'm not sure what we can do about the select functionality. I was hoping that you would know where to add code to handle that. Did you make any progress after you posted your last comment above?

I would like to start using the feature introduced in this pull request and the bug fixes from #674 (sorry for messing up the merge there by targeting my pull request against the wrong branch!). Are you ready to cut a release with these two pull requests in it, or is anything in your way that I could help with?

josdejong:
@AdamVig I will do a release now. I haven't worked on the `select` issue since my first short try. Since this is a new feature and no regression I'm ok with doing a release now. We can fix/improve further in a bug fix release.

josdejong:
@AdamVig the defaults styling is now available in `v5.32.0`. Thanks again :)

josdejong:
I've tried again to get the classes working on enums, and now it worked right out of the box with the exact same code [I posted before](https://github.com/josdejong/jsoneditor/pull/666#issuecomment-473675457). I was probably testing with a JSON Schema not having a default value on the enum I tested or something like that :joy: . See b279ed3069b9b4650668595071bfe58e29b5e8c9.

josdejong:
Fixed now in `v5.32.2`

PR #670, opened 2019-03-12
fix(i18n): fix Chinese translation error

yuxizhe:
fix Chinese translation error,
` 重复键` is noun ,  `复制键`is verb


josdejong:
Thanks @yuxizhe , I believe you :)

PR #671, opened 2019-03-12
fix: when duplicate keys occur discard get function

yuxizhe:
when duplicate keys occur, onChangeJSON get a wrong json. when using in react and mobx, this bug will cause new key-value cover the original key-value
bug-json

### DIFF
--- a/src/js/treemode.js
+++ b/src/js/treemode.js
@@ -291,6 +291,14 @@ treemode.get = function () {
     }
   }
 
+  // discard when duplicate keys
+  if (this.node) {
+    var duplicateErrors = this.node.validate();
+    if (duplicateErrors && duplicateErrors.length > 0) {
+      throw duplicateErrors[0];
+    }
+  }
+
   if (this.node) {
     return this.node.getValue();
   }


josdejong:
Thanks @yuxizhe this is an interesting case. 

I'm thinking about how to solve this best. Shouldn't we move this theck for duplicate errors in the `JSONEditor.get()` method instead, and throw an exception there?

yuxizhe:
@josdejong 
maybe like this?
```js
/**
 * Get JSON object from editor
 * @return {Object | undefined} json
 */
treemode.get = function () {
  // remove focus from currently edited node
  if (this.focusTarget) {
    var node = Node.getNodeFromTarget(this.focusTarget);
    if (node) {
      node.blur();
    }
  }

  // discard changes when duplicate keys
  if (this.node) {
    var duplicateErrors = this.node.validate();
    if (duplicateErrors && duplicateErrors.length > 0) {
      throw duplicateErrors
    }
  }

  if (this.node) {
    return this.node.getValue();
  }
  else {
    return undefined;
  }
};
```

josdejong:
Yes, something like that. Maybe throw just the first error in case there are multiple.

It may make sense to think this through a bit more. Just thinking aloud here:

So far, for the `code` mode you can get an error quite easily whilst you're typing and try to get the JSON with `editor.get()`, because the text cannot always be parsed. `editor.getText()` does always work and never throws an error. When there is invalid JSON, you clearly see where the error is with a red error icon in the gutter:

code_mode_error

For mode `tree`, calling `editor.get()` always works except one special case: when a changed value contains an invalid unicode character like `foo\A0`. Both `.get()` and `.getText()` will not work in that case. It makes sense to me to throw an error too when there are duplicate keys, since that cannot be represented in JSON, which can lead to the issue you bring up. To make the behavior more consistent over all, I think it's an good idea to change the yellow warning sign in case of a duplicate property with a red error icon too. And ideally we display that too in case of invalid contents of a value. What would be really nice but complicated to achieve is to make `editor.getText()` work too when there is an error.

So, to summarize, I think your PR is a great first step, and after that we can look into improving error icons.

yuxizhe:
@josdejong  thanks for your explain, its a pleasure to participate in improving error icons.

josdejong:
:+1: let's do one step at a time. Can you adjust this PR to throw an exception in `JSONEditor.get()` in case of duplicate properties? In a separate PR we can adjust the icon of duplicates errors etc.

josdejong:
:joy: I see you already updated the PR, I will do some testing with it.

yuxizhe:
@josdejong everything goes well?  ;)

josdejong:
Yes :) Your PR is ready to be merged but I want to think this though a bit more. I think it is best to release this after we've changed the icon, and I had one more idea: when duplicating a property, change the field to a new, unique name (like `name` => `name (copy)` or `name (copy 2)`, etc), so duplicate fields will occur less often.

yuxizhe:
@josdejong wow, good idea! 
 But when user just input the name,  this bug still happen.

josdejong:
> But when user just input the name, this bug still happen

Yes indeed, it will not let the issue disappear but I think it will occur much less often.

josdejong:
@yuxizhe I've merged your PR in the `duplicate_key_errors` branch, and I've added a change to display duplicates as errors, and when duplicating a new unique field name will be created.

duplicates

josdejong:
@yuxizhe I was working out error icons for invalid text (like an invalid unicode) similar to duplicates. There are some more complications with these duplicate and invalid text errors, like undo/redo. 

Then I was thinking in a totally different direction to solve this issue: simply keep these issues local and only accept them whilst typing. When leaving a field, have the editor either fix the issue (change name into a unique name in case of duplicates) or revert to the last valid value (in case of invalid unicodes). It means that `editor.get()` can never throw an exception which simplifies things a lot. Will try out this approach.

josdejong:
I've a solution working in branch `duplicate_key_errors`, if you want you can have a look @yuxizhe to get a feel how it works. 

Still need to fix one bug in rendering the warning icon in case of duplicate fields.

yuxizhe:
Great! 👍 I will try to keep up with your progress.

josdejong:
I've fixed the last issues that I saw, and tested with a React demo project to verify that the original issue is resolved. I think it's ready to merge in `develop` and publish with the first next release. 

I would love to have some feedback from you @yuxizhe about this approach, would be nice if you could give it a try. You can try it out by checkout out the [`duplicate_key_errors_take2`](https://github.com/josdejong/jsoneditor/tree/duplicate_key_errors_take2) branch.

yuxizhe:
@josdejong I test in my project, it works perfectly.

image

What about replace spaces with underscore `_` ?   Because  space would become` %20` in url.

eg
`symbol (copy 6)` then  `symbol%20(copy%206)`

I think `symbol_copy_6` would be better :)

josdejong:
Thanks for your feedback :+1: will merge this solution in `develop` then.

In general, an object property has no relation with a url (in your case it does?). I used `symbol (copy)` so it looks human friendly. Normally, you want to replace the `(copy)` part with something else anyway, it's just a mechanism to prevent duplicate names.

josdejong:
I've published the new way of handling duplicates now in `v5.32.2` :tada:

PR #674, opened 2019-03-15
Fix Node._findSchema bugs with references and missing paths

### LINKED ISSUE #667: JSON Schema examples not displayed for referenced schemas
When adding some `examples` to the JSON schema, they are displayed when hovering over the field name. This seems not to work for nested schema's, loaded via reference. 

@AdamVig am I overlooking something, should this work for referenced schema's?

Also it looks like the fields display the description of the object they belong to (hover over `job.salary` in the following example shows "Job description"), I'm not sure if that is the intention.

This can be reproduced by adding some `examples` fields to the JSON schemas in  `07_json_schema_validation.html`, demo:  https://jsbin.com/nohohixuya/edit?html,output

AdamVig:
Fixes #667.

### DIFF
--- a/examples/07_json_schema_validation.html
+++ b/examples/07_json_schema_validation.html
@@ -49,7 +49,8 @@
       "age": {
         "description": "Age in years",
         "type": "integer",
-        "minimum": 0
+        "minimum": 0,
+        "examples": [28, 32]
       },
       "job": {
         "$ref": "job"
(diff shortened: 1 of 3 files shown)


josdejong:
Thanks a lot Adam, works like a charm now!

josdejong:
I've merged this PR in `develop` manually but forgot that this PR still pointed to `master`. It will be marked as merged as soon as we merge everything to `master` with the first next release.

josdejong:
@AdamVig your fix is now published in `v5.32.0`. Thanks!

PR #679, opened 2019-03-28
Fix JSON path parsing for array indices

AdamVig:
- Coerce numeric indices to numbers in parsePath
  This fixes a regression introduced in 3e7e1cebfd237ed7a41f5f24cabd84b6db8fd625.

- Switch JSON path tests to strict equality
  The tests for `stringifyPath` and `parsePath` were using
  [`assert.deepEqual`](https://nodejs.org/api/assert.html#assert_assert_deepequal_actual_expected_message), which was causing the tests to pass when they should not have. Beyond that, the `deepEqual` method is deprecated. The suggested replacement is [`assert.deepStrictEqual`](https://nodejs.org/api/assert.html#assert_assert_deepstrictequal_actual_expected_message), which causes the tests to fail where they should.
  
  The difference between the two methods is the different between `==` and `===`. `deepEqual` would coerce the input `"2"` to `2`, so the test was passing even though the output was of the wrong type (string instead of number).

### DIFF
--- a/src/js/util.js
+++ b/src/js/util.js
@@ -797,6 +797,8 @@ exports.parsePath = function parsePath(jsonPath) {
         if (index.length === 0) {
           throw new Error('Invalid JSON path: array value expected at index ' + i)
         }
+        // Coerce numeric indices to numbers, but ignore star
+        index = index === '*' ? index : JSON.parse(index);
         path.push(index);
       }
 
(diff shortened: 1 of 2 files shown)


josdejong:
Thanks Adam for fixing this regression!

AdamVig:
Hey @josdejong, sure thing! When do you think it will be released?

josdejong:
Will do right now to make it your lucky day ;)

josdejong:
Should be fixed now in `v5.32.1`

AdamVig:
Thanks for the quick turnaround!

PR #685, opened 2019-04-05
Feat: Added filter and trigger options for autocomplete

Gcaufy:
**filter**: I need autocomplete work with `contain` rule instead of `startWith`
**trigger**: I need to trigger autocomplete when I focus instead of type something in it. then user can simply choose an option.


josdejong:
Thanks @Gcaufy , great idea to make the autocomplete more flexible in this regard. I will review your work in a couple of days.

josdejong:
Your PR looks good @Gcaufy , thanks :+1: ! The `filter` option is very powerful since you can now pass your own function if you want :)

A few feedback points:

1. Can you please replace `const`, `let`, and object getters with old fashioned `var` and regular properties? We don't have an ES5 transpiler in place yet :( I've opend an issue for it, it's getting more and more painful: #689.
2. A minor glitch: in the updated 12_autocomplete_dynamic.html, click "Insert" in the action menu, then the (empty) property input has the focus. Now, when clicking on the (empty) value input, it should get the focus but it doesn't. The other way around works fine.

josdejong:
@Gcaufy it would be great if you can address the two feedback points so we can merge your work.

Gcaufy:
Move to: https://github.com/josdejong/jsoneditor/pull/713

PR #707, opened 2019-05-29
Fix #697 JSON Schema enum dropdown not working inside an array

### LINKED ISSUE #697: JSON Schema enum dropdown not working inside an array
See this jsbin: https://jsbin.com/fopotoh/edit?html,output

See also https://github.com/josdejong/jsoneditor/issues/391#issuecomment-491613052

josdejong:
@AdamVig Can you please check whether this fix makes sense and check if I'm not overlooking something obvious?

### DIFF
--- a/test/Node.test.js
+++ b/test/Node.test.js
@@ -27,6 +27,33 @@ describe('Node', function () {
             assert.strictEqual(Node._findSchema(schema, {}, path), schema.properties.child);
         });
 
+        it('should find schema inside an array item', function () {
+            var schema = {
+                properties: {
+                    job: {
+                        type: 'array',
+                        items: {
+                            properties: {
+                                company: {
+                                    enum: ['test1', 'test2']
+                                }
+                            }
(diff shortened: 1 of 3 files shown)


test/Node.test.js
```diff
@@ -27,6 +27,33 @@ describe('Node', function () {
             assert.strictEqual(Node._findSchema(schema, {}, path), schema.properties.child);
         });
 
+        it('should find schema inside an array item', function () {
+            var schema = {
+                properties: {
+                    job: {
+                        type: 'array',
+                        items: {
+                            properties: {
```
AdamVig: You might want a `type: 'object'` above this line to make this valid JSON schema. Just for clarity.

josdejong: :+1:

josdejong:
Thanks Adam. I'm really glad you started setting up unit tests a while ago! Was very easy to extend now :)

PR #712, opened 2019-06-06
fix(css): remove :hover style on disabled button.

Gcaufy:
Should not have :hover style on disabled buttons.


josdejong:
Thanks for your PR @Gcaufy, makes sense .

PR #713, opened 2019-06-06
Feat: Added filter and trigger options for autocomplete 

Gcaufy:
filter: I need autocomplete work with contain rule instead of startWith
trigger: I need to trigger autocomplete when I focus instead of type something in it. then user can simply choose an option.


josdejong:
Thanks @Gcaufy, looks good :+1:

josdejong:
Published now in `v5.34.0`

Issue #763, opened 2019-08-07
Cannot read property 'getPath' of undefined


Thaina:
https://github.com/josdejong/jsoneditor/blob/3f182a1f0411764fabab1d9bb35bbfbaaa41f67c/src/js/treemode.js#L1496-L1499

If we set `autocomplete.trigger` as `focus` it will go to line 1498 but then `node.getPath()` will thrown if it cannot find node when we press it here

![image](https://user-images.githubusercontent.com/1042507/62611903-ae93c780-b930-11e9-9e4a-4b9f71bf1170.png)

Should we check `node` variable and return when it was not a Node?

josdejong:
Fixed now in `v6.3.0`

PR #764, opened 2019-08-07
for #763

Thaina:
#763

### DIFF
--- a/src/js/treemode.js
+++ b/src/js/treemode.js
@@ -1493,32 +1493,34 @@ treemode._showAutoComplete = function (element) {
 
   var self = this;
 
-  setTimeout(function () {
-      if (self.options.autocomplete.trigger === 'focus' || element.innerText.length > 0) {
-          var result = self.options.autocomplete.getOptions(element.innerText, node.getPath(), jsonElementType, node.editor);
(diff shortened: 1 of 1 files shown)


josdejong:
Thanks for this fix @Thaina . Can you please change the indentation to 2 space?

Thaina:
@josdejong Right now in workplace I don't have access to the repo and ssh key from my home machine. Couldn't you just pull and change it? Or actually you could just copied mine and push in your repo then drop this pull request. I just want this to be fixed

josdejong:
Yes will do, thanks for your quick reply.

I will also do some deeper digging, I'm not sure whether this solves the cause, maybe node should not be undefined.

Thaina:
@josdejong I have dig the code and see that `node` are `undefined` because the dropdown menu for insert/delete the node is not a node itself yet when it clicked it still invoke the autocomplete

The actual fix might be that, the dropdown menu should not invoke autoComplete. because it cannot typing anything there

josdejong:
yes indeed. I've added an extra check before calling `_showAutoComplete`, now and also made itself `_showAutoComplete` robust against missing node (like your PR).

Issue #760, opened 2019-08-06
 Problem with css code for overriding  


ppetkow:
[div.jsoneditor ](https://github.com/josdejong/jsoneditor/blob/develop/dist/jsoneditor.css)

Hello, i found this as hough problem. 

Your code start really strange for your application. Our devs used a lot, but we need to change style of this app and hit this css problem. 1st is css not some preprocessor like scss or less (prefer scss). 

2nd in your css code on 4th row you make 2 important in beginning of file!! I think this is to much and is really incorrect. Can you improvе option for overriding by removing !important's and start all styles from one general class name only. 

`<div  class="jsoneditor"> ... </div>`  

And all inside this div depend on this class name. 

Also i strongly suggest you to move it on SCSS. Then the project will be more flexible and good.

josdejong:
Thanks for your input. 

The usage of `!important` is indeed not pretty, though you can still override it for custom styling. 

A PR replacing the usage of `!important` or even move to SCSS would be welcome :)

ppetkow:
Ok i will update project :) and move to SCSS

ppetkow:
More Q :) 

Search box is too complicate with this 2 tables, can i update it and move it to divs. This means also to change display to flex. But you write on website that  the tool support IE8+, but if i want to use flex-box model needs to pump version of IE to 10. If this was compromise from your side i will be happy to upgrade. :)

josdejong:
Thanks for your offer! The IE8+ was a leftover, I've changed that to IE11 now (like JSONEdtior itself already mentioned in the readme: https://github.com/josdejong/jsoneditor#json-editor). 

A small refactoring to use flexbox here and there would be very welcome! That's also needed for the main menu / navigation bar / contents / status bar. Please be careful not to make a _too_ big PR at once, preferrably one PR addresses one thing, else it's hard to review.

ppetkow:
> A small refactoring to use flexbox here and there would be very welcome! That's also needed for the main menu / navigation bar / contents / status bar. Please be careful not to make a _too_ big PR at once, preferrably one PR addresses one thing, else it's hard to review.

know ... to late .. movement from css to scss was big changes for files

josdejong:
@ppetkow is your initial concern resolved with the refactoring to SCSS that you did? Can we close this issue?

ppetkow:
Yes, definitely! Thank you

josdejong:
:+1:

PR #765, opened 2019-08-07
Problem with css code for overriding #760

ppetkow:
- There is a lot of changes on css files, but this request affect and 2 js file for modal setting and change GULP configuration to support scss compilation. 
- The naming and file structure in dist folder is same.
- I keep all visual styling of app. 
- Move all colors to variables for 3th part easy overriding 
- Remove some !important properties


josdejong:
Thanks @ppetkow , this looks really good! I'm really glad with your effort to pick this up, it's a great improvement. And sorry for my late reply, it's holiday season.

Some feedbacks:

1. great that you pulled out colors and styling into `styles.scss`!
2. shouldn't `jsoneditor.scss` and all others import `styles.scss` instead of the other way around? `jsoneditor.scss` depends on `styles.scss` and not the other way around right?
3. I noticed one small glitch: the text color of the search result summary left from the search box (like "5 results") should be white but is black.
4. My fingers are itching to refactor the main structure (menu, nav bar, contents, errors, status bar) with flexbox but probably better to address that later, the PR is already quite large.

ppetkow:
No problem for speed :) Im on work :D I will check this points and make new commit. 

Yep number 4 is for next PR maybe ;)

josdejong:
Your latest changes looks good :+1: will merge your work now.

josdejong:
@ppetkow FYI: you're SASS implementation is now published in `v6.4.0` (the improved search box too). Thanks again :)

PR #782, opened 2019-09-03
fix：Modify the file address reference method

moonbreezee:
I think this should be consistent with the other file references (searchbox.scss, jsoneditor.scss), otherwise the jsoneditor-icons.svg file will not be found after the file-loader processing in webpack.


josdejong:
Thanks for this fix @moonbreezee

PR #809, opened 2019-10-08
Fix #727

### LINKED ISSUE #727: Emit the focus and blur event?
Hi,
     Can you support emitting the focus and blur event?

123survesh:
Have exposed the **onFocus** and **onBlur** events through the **JSONEditor.options**


josdejong:
Thanks Survesh! I'll do testing and review of your PR coming weekend.

josdejong:
@123survesh I've been doing some testing. The focus/blur works great in tree mode for input fields in tree mode :+1: 

Feedbacks:

1. I think it is necessary though to make it work for other modes too (text, code, preview). 
2. I also noticed that focus/blur currently doesn't work with the search bar and buttons of the main menu, and also not when the context menu is opened. 
3. When switching focus from one input to the next (by clicking an other input or clicking Tab), the editor emits a `blur` and then a `focus` again. I understand why this is, but ideally these two events should not fire: the editor still has focus. I'm not sure if this is technically easy to realize, and this may not be a must but a nice to have. What do you think?

123survesh:
I have deviated from the actual need, that's why the focus and blur events are firing only for the input fields. Instead of going for the whole editor I went for the individual nodes I guess.

I will try to cover all the use cases where the editor as a whole is active, this should prevent point 3 as well.

josdejong:
Sounds good, thanks!

123survesh:
Hi @josdejong , 

I tried to track the focus of the editor with the browser's "focus" event by using the attrbute **tabindex** that did not work well. So I added a **click**  event listener on the **document** object and I am tracking if the clicks are on/in the editor or out. This covers the points that you had mentioned in your previous comment. 

Do check and tell if the approach I have taken for this issue now is ok.

josdejong:
Thanks, will have a look at it soon Survesh, I expect Wednesday (also your other open PR on custom value rendering).

123survesh:
Ok @josdejong.

josdejong:
This is a really nice solution Survesh, it works very smoothly! I see you neatly create and cleanup the `document` listener.

1. Currently there are three almost equal implementations of `focusEventTracker`. Would it be possible to extract this function in a separate file and reuse it three times? I'm afraid it will be hard to maintain otherwise.
2. Currently the blur/focus listens for mouse clicks. It would be nice to extend this to also listen for keyboard events: currently you can use `Tab` to move the focus from some button above the editor to buttons inside the editor without the focus/blur monitor noticing.

123survesh:
Hey @josdejong ,
Yes it should be possible to put the code in a separate file and use them across those three cases. I will see how to do this and track the tab button presses as well.

123survesh:
Hey @josdejong , I have separated the` focusEventTracker()` from last commit into a `FocusTracker` class and have used it in the three available modes now. I have also added a Event Listener that tracks the press of 'Tab' key.

After implementing the 'Tab' key tracker, I noticed that the focus event did not propagate out of the editor after it reached the last element in the editor. So I added a patch to fix that 8c631dc.

josdejong:
Looks really good @123survesh, thanks! I will do some functional testing in a couple of days.

Two small ideas when looking at the code:

1. Shouldn't the contents of the method `.add()` of the `FocusTracker` be moved to its constructor? I don't think a separate method call is really needed?
2. For consistency it may be a good idea to rename `.remove()` to `.destroy()` (most components have such a method)

123survesh:
>     1. Shouldn't the contents of the method `.add()` of the `FocusTracker` be moved to its constructor?
I was thinking of a use case where people will need to add and remove the event listeners more than once. But we only use it once here now, so I will move the `add()` method inside the constructor.

>     2. For consistency it may be a good idea to rename `.remove()` to `.destroy()` (most components have such a method)
Yep will do :+1:

src/js/FocusTracker.js
```diff
@@ -0,0 +1,99 @@
...
+
+    this.focusFlag = false
+    this.trackerSetFlag = false
+    this.firstEventFlag = true
+
+    /*
+      Adds required (click and keyup) event listeners to the 'document' object
+      to track the focus of the given 'target'
+     */
+    if (!this.trackerSetFlag) {
```
josdejong: I think this `if (!this.trackerSetFlag)` is redundant now :) 

Also, I don't think `FocusTracker` is useful at all without `this.target`, so that can probably be just required instead of optional?

123survesh: Ah yeah my bad :sweat_smile:, I will remove the dependency on the `this.trackerSetFlag` . Also to make `this.target` required I am adding a check for its availability and throwing an error if its not there.

josdejong:
Code looks good @123survesh , and I've been testing more thoroughly and it works like a charm, the file `test_focus_tracker.html` with the fat red borders makes that really easy.

I found only one edge case, maybe you can see if that is easy to solve. If not I think it's fine to just leave that as it is now, it's really an edge case. To reproduce: 
- click the context menu of a node in the editor (it will get focus if it doesn't yet have focus). 
- Then click "Insert". Now a new field will be inserted and gets focus, but the `onBlur` event is triggered and the red border around the editor disappears.

123survesh:
Ok, yes. I checked it out. I faced this issue when changing modes using the mode changer button as well.
The `FocusTracker` I wrote only checks if the current click event target is either the editor or is within the editor at the time of event. But the popup menus  get destroyed once they are clicked upon. And hence the `FocusTracker` is not able to pick them up before they get destroyed.
Do you have any ideas on how to tackle this?

josdejong:
Makes sense, I understand why that happens. I'll give it a try see if I can come up with a solution. Will be coming weekend I'm afraid though, have some busy days ahead.

josdejong:
I've done some testing and trying. First I tried with the HTML events `focus` and `blur`, but that doesn't work in all cases. 

What could work is creating a method `FocusTracker.focus()`, and calling this from all places that we know will give the editor focus. The treemode, textmode, and previewmode all have a `.focus()` method already, and `Node` too. And we could extend the `.focus` method of `ModeSwitcher` to also call `FocusTracker.focus()`. This is a bit of a tricky solution though which requires calling this focus method carefully in all cases. What do you think?

123survesh:
Yep that is ok with me. Not all cases, where the focus() method is being used can be tracked and we will need a override in the FocusTracker to set it manually. 

I have added another small condition to widen the check for focus. With this some of the focus events that were missed before (like inserts, mode switching) are handled now. I missed this before. 

But this check won't work if the FocusTracker's checking code gets control before the custom focus() method that is being used. If the code for focus() method always gets executed before the FocusTracker's check code, we do not have to add a manual override for the FocusTracker.

josdejong:
Wow, that's an elegant solution and it works like a charm! I'm glad there is no need for special checks all over the place.

One last minor detail: currently, when the editor has focus and you switch to an other mode, the editor is destroyed and a new one is created, which immediately fires a focus event again. So, you see a focus event twice in the console. Would it make sense for consistency to fire a `blur` event when the editor has focus and `FocusTracker` and is destroyed?

123survesh:
Yeah you are right, it should. Sorry that I keep on missing features like this :sweat_smile:

123survesh:
I will update with a PR

josdejong:
> Yeah you are right, it should. Sorry that I keep on missing features like this sweat_smile

No problem at all!

josdejong:
Nice :+1: ! Will merge your work now Survesh, thanks again. I will do a release somewhere this week.

josdejong:
The new `onFocus` and `onBlur` are now published in `v7.3.0`, thanks again Survesh!

PR #813, opened 2019-10-10
some portuguese translations added

victorananias:
I translated to portuguese everything that was pending. The ones that I believed that couldn't be the best translation I kept the TODO warning.


josdejong:
Thanks a lot Victor!

PR #829, opened 2019-10-24
fix the z-index for context button in tree mode

yeahnoob:
Try to fix the following problem related to CSS `z-index`

Screenshot_20191024_152333

Screenshot_20191024_152413

### DIFF
--- a/src/scss/jsoneditor.scss
+++ b/src/scss/jsoneditor.scss
@@ -144,6 +144,7 @@ div {
         margin: 0;
         border: none;
         cursor: pointer;
+        z-index: auto;
         background: transparent url("./img/jsoneditor-icons.svg");
         &:focus {
           background-color: $jse-preview;


josdejong:
Thanks Xanadu for your fix, it indeed fixes the issue but there was something more going on. 

I had a bit deeper look into where this issue with `z-index` is coming from, and it turns out there was a naming collision in css: both the dropdown menus and the context menu button used the css name `jsoneditor-contextmenu`. I've renamed the css class of the button to `jsoneditor-contextmenu-button`, and also go over all `z-index` usages to see if they are still relevant:

- context menu
- popover (JSON schema)
- dropdowns in the Transform modal (Selectr)
- auto completion
- tree path dropdown (show all)

Some where not relevant anymore so this cleans up quite a bit, see this commit: 009d896c30427f8eebb5e6773f4f580d1483a1ac.

josdejong:
Fixed now in `v7.2.1`

PR #839, opened 2019-11-02
onNodeName() new param current-node-value as value

sentinelburo:
Id like to use this as follows:

```js
    onNodeName(param) {
      const {path, type, size, node} = param;
      if (type === 'object' && size > 0) {
        console.log(param);
        if (node.childs[0].field === 'name') return node.childs[0].value;
      }
    }
```

### DIFF
--- a/src/js/Node.js
+++ b/src/js/Node.js
@@ -3943,7 +3943,8 @@ export class Node {
           nodeName = this.editor.options.onNodeName({
             path: this.getPath(),
             size: count,
-            type: this.type
+            type: this.type,
+            node: this
           })
         } catch (err) {
           console.error('Error in onNodeName callback: ', err)


josdejong:
Thanks for your PR @sentinelburo, I understand what you try to achieve. 

So far, the API doesn't expose the internal `Node` objects and tries to keep the API simple. What you can do though is something like the following (code not tested, I hope you get the gist):

```js
import { get } from 'lodash'

// ...

    onNodeName({ path, type, size }) {
      const data= get(myEditor.get(), path)
      if (type === 'object' && size > 0) {
        if ('name' in data) {
          return data.name
        }
      }
    }
```

Would that work for you?

sentinelburo:
Yes this would work .

but editor is undefined since we need the options to create the instance.
in my Setup, import { get } from 'lodash' does not work native in the chrome browser, 

let editor;

const options={
onNodeName({ path, type, size }) {
      const data= get(editor.get(), path)
      if (type === 'object' && size > 0) {
        if ('name' in data) {
          return data.name
        }
      }
    }
}

editor = new Editor(el, options);

josdejong:
> but editor is undefined since we need the options to create the instance.

Ah, that's a good point. How about passing `editor` as a property to `onNodeName`? Like:

```js
onNodeName({ path, type, size, editor }) {

}
```

> in my Setup, import { get } from 'lodash' does not work native in the chrome browser

yeah, so that would require either loading lodash, or using your own util function to check get contents of the path. I don't think that is a big issue though.

sentinelburo:
you like simple api (no need to change it later)
1.) i need just the object ( same object i would get by the call to any of the previous code).
2.) giving back the object would be good for api too couse its the raw data given by the user
3.) giving back the editor or any other internal object would lead to ppl use internal function, which leads to the point where it is hard to change things later without breaking code.

.) its not too hard to use lodash in chrome with import() or any other lib :)

josdejong:
yes, that's an interesting idea, passing the current value too, so it would be like:

```js
onNodeName({ path, type, size, value }) {

}
```
There could be a small performance penalty since you have to get and pass the json value every time again, also in cases where you do not actually use the `value`, but that may be negligible.

I think it cannot do harm to pass the `editor` instance itself: it's simply what you get when creating an editor `const editor = new JSONEditor(...)`, this is already public API. It would solve the chicken-egg problem you explained before. The internal `Node` though is not and I would like to keep that non-public.

josdejong:
@sentinelburo what do you think? Shall we go for passing `value` as new argument, or `editor`, or both?

sentinelburo:
I would opt for the value as property.
cause its not sure if editor has a good inner state while it is rendering its layout.
it may lead to bugs if some functions are used before the rendering is finished.

josdejong:
Sounds good, lets start simple then and just implement `value`. Can you give that a try or do you want me to pick this up?

sentinelburo:
i d like to update this pr with the value as new prop.

josdejong:
Thanks @sentinelburo :+1: I will await an update in your PR then (or a new PR, whatever works easiest).

sentinelburo:
how should we name the param (value, current_value, ...)

josdejong:
I think we should call it `value`, that is consistent with other callbacks that JSONEditor has (check out the docs to get an idea).

josdejong:
Closing this PR in favor of #1409

Issue #612, opened 2018-12-10
Implement onValidationError option and a public validate API method


josdejong:
It would be nice to have a callback function and a method to run/listen for validation errors. This can be used for example to check whether a document is valid before submitting it, or to render validation errors on a different or more prominent place in a web application.

We should think about how this API should look exactly: think about the how to treat both JSON parse errors, JSON Schema validation errors, and custom validation errors.

Followup from discussion here: https://github.com/josdejong/jsoneditor/pull/560#issuecomment-445551510

josdejong:
If anyone is interested in implementing this feature please let me know!

optimalisatie:
We will have a look into it but I am not certain yet that we can provide a solution on short term. 

For anyone interested, the following quick fix is what we are currently using. It requires to load [ajv](https://github.com/epoberezkin/ajv) seperately from json-editor.

```javascript
// load ajv (JSON schema validator)
var AJV = new window.Ajv({
    allErrors: true,
    verbose: true
});

// add a JSON schema to ajv
// schemaSrc = JSON schema object
function add_schema(name, schemaSrc) {
    AJV.addSchema(schemaSrc, name);
    return AJV.getSchema(name).schema;
}

// validate json against schema
function validate_schema(name, json) {
    return AJV.validate(name, json) || AJV.errorsText(); // ajv can return more details, the .errorsText() method is a easy shortcut that serves any error(s)
}

// onChange event in JSON editor options
new JSONEditor('container', {
    /* ... */
    onChange: function() {

        try {
            var json = critical_json_editor.get();
        } catch (e) {
            return;
        }

        var errors = validate_schema(name, json);
        if (typeof errors === 'text') { /* process json errors */ }
    }
}
```

The disadvantage of the solution is that AJV will run 2x and requires additional parsing. An `onValidationError` callback would provide better performance.

josdejong:
:+1: thanks for sharing this workaround Jan Jaap.

as a workaround, I think you could also do:

```js
var errors = myJsonEditor.validateSchema(json)
```

instead of setting up AJV yourself. No official API but should work for now as a workaround.

tscislo:
+1 For this feature

JohnBDonner:
@josdejong I don't think it's possible to reference the jsonEditor object in this scenario, since it doesn't seem possible to edit the options after initialization:

```javascript
var jsonEditor = new JSONEditor('container', {
    /* ... */
    onChangeJson: function(json) {
[...]

PR #854, opened 2019-11-29
onValidationError option

meirotstein:
This PR adds the `onValidationError` option as discussed on #612

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -162,6 +162,28 @@ Constructs a new JSONEditor.
   ```
   
   Also see the option `schema` for JSON schema validation.
+
+- `{function} onValidationError(errors)`
+
+  Set a callback function for validation errors. Available in all modes.
+
+  On validation of the json, if errors were found this callback is invoked with the validation errors data.
+  Between validations, the callback will be invoked only if the validation errors were changed.
+  
+  Example:
+
+  ```js
+  var options = {
+    /**
+    * @param {Array} errors validation errors
+    */
+    onValidationError: function (errors) {
+      ...
+    }
+  }
+  ```
+  
+  Also see the option `schema` for JSON schema validation.
   
 - `{function} onCreateMenu(items, node)`
   
--- a/docs/api.md
+++ b/docs/api.md
@@ -169,7 +169,7 @@ Constructs a new JSONEditor.
 
   On validation of the json, if errors were found this callback is invoked with the validation errors data.
   Between validations, the callback will be invoked only if the validation errors were changed.
-  
+
   Example:
 
   ```js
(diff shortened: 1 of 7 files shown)


josdejong:
Thanks Meir, this works like a charm! I see you created a clear example and wrote docs for the new feature :+1: .

To small feedbacks:

1. The custom onValidate error in the example (`'Member must be an object with properties "name" and "age"'`) did did confuse me, I thought there was an error in the editor. Maybe replace this custom error with something realistic like actually check whether the age is between 18 and 99 or something like that?
2. I understand the function `isValidationErrorChanged`, but I think we could make this a bit more generic: It basically checks whether two objects are equal, right? Can we replace this with say the `isEqual` of lodash? (we're not yet using lodash but there are quite some places where we could use it). Or we could use a simple `JSON.stringify(currErr) === JSON.stringify(prevErr)`. Or are there specific reasons we cannot simply compare the two arrays as a whole?

meirotstein:
**Custom validation example**: you are right - a bad copy of another example :grimacing: - will be fixed.
**`isValidationErrorChanged` implementation**: two reasons:
1. I first tried the `JSON.stringify` solution but it didn't work, I guess there were minor changes between the error objects that I cannot explain, anyway I didn't get deep into it because of the 2nd reason:
2. I feel a bit uncomfortable with `stringify` because I'm afraid that on very big documents with many errors it can get quite heavy (serialization + comparison).
The current approach is more gradual, and if there's a difference somewhere in the middle it will not have to go over the entire arrays.
I didn't try the `lodash` simply because this lib doesn't have this dependency - there's a usage of it on the React examples, but not on the main bundle. I really like `lodash` and use it on many other projects and for sure will try this option if it's available.

josdejong:
I think you're right that with larger objects and many errors stringify will probably not perform well (though that's an assumption ;) ). 

I think your current solution is fine, just looking if we can simplify it somehow. We can just add `lodash` to the project if that helps us, but if stringify doesn't work, the lodash `isEqual` will probably also not work? Would be nice if you can give the `isEqual` approach a short try and if it doesn't work out we can just keep the current solution.

meirotstein:
@josdejong I will give it a try

meirotstein:
I have tested it - it will not work 😄 from the  same reason that `stringify` is not working:
The validation error object also contains the changed value - that makes the previous and current object always differ.

josdejong:
Ahh, of course, that makes sense :smile: . So then it would only work with a customized comparator, which is what you've already built with `isValidationErrorChanged`.

Thanks for updating the example. Will merge now

josdejong:
`onValidationError` is now available in `v7.4.0`, thanks again!

PR #863, opened 2019-12-06
add timestampFormat

smallp:


josdejong:
Thanks @smallp! I will review and test your PR this weekend.

src/js/Node.js
```diff
@@ -1791,8 +1791,20 @@ export class Node {
           this.dom.value.parentNode.appendChild(this.dom.date)
         }
 
-        this.dom.date.innerHTML = new Date(value).toISOString()
-        this.dom.date.title = new Date(value).toString()
+        let title = false
```
josdejong: Is it more neat to define `let title = null` instead of `false`?

src/js/Node.js
```diff
@@ -1791,8 +1791,20 @@ export class Node {
...
+            field: this.field,
+            value: this.value,
+            path: this.getPath()
+          })
+        }
+        if (!title) {
+          this.dom.date.innerHTML = new Date(value).toISOString()
+          this.dom.date.title = new Date(value).toString()
+        } else {
+          this.dom.date.innerHTML = title
```
josdejong: Since this is a public API now, I think it makes sense to replace using innerHTML:
```js
this.dom.date.innerHTML = title
```

 with something more strict like:
```js
while (this.dom.date.firstChild) {
  this.dom.date.removeChild(this.dom.date.firstChild)
}
this.dom.date.appendChild(document.createTextNode(title))
```

josdejong:
Looks good and works like a charm Small! Thanks.

I'm thinking about one API detail: the default implementation of the timestamp tag sets both the text contents of the tag, and a title: 
```js
this.dom.date.innerHTML = new Date(value).toISOString()
this.dom.date.title = new Date(value).toString(
```
The API of `timestampFormat` currently doesn't allow setting the title, it only allows setting the innerHTML. That feels a bit inconsistent. We could simply remove the existing `.title = ...`, or change the API to allow passing both tag text and title. What do you think?

smallp:
Yes, `let title = null` is better, and we should use `createTextNode` to avoid secure issue.
As for the `title`, it is a litter bit strange if `timestampFormat` retrun an object. We could simply remove the existing `.title = ...`

josdejong:
Thanks for the updates. Yeah maybe we shouldn't overengineer the `timestampFormat` option at this point. 

I was just thinking about an other option: keep the current `this.dom.date.title = new Date(value).toString()` and simply apply it too in case of a custom timestamp format. I use the title myself very regularly because it gives me local time.

src/js/Node.js
```diff
@@ -1791,8 +1791,22 @@ export class Node {
           this.dom.value.parentNode.appendChild(this.dom.date)
         }
 
-        this.dom.date.innerHTML = new Date(value).toISOString()
-        this.dom.date.title = new Date(value).toString()
+        let title = null
```
josdejong: I try to use `const` instead of `let` as much as possible, it often simplifies the code. In this case it is possible to do something like:

```js
const title = (typeof this.editor.options.timestampFormat === 'function') 
  ? this.editor.options.timestampFormat({ ... })
  : new Date(value).toISOString()

// ... no need for the if (!title) {... } else { ... }
while (this.dom.date.firstChild) {
  this.dom.date.removeChild(this.dom.date.firstChild)
}
this.dom.date.appendChild(document.createTextNode(title))
```

That is less if/elses, and also clearly splits formatting logic and rendering logic.

smallp: `if (!title) {... } else { ... }` is necessary. If `timestampFormat` return null, we need to fallback to the default setting.

smallp:
I think keeping the current `this.dom.date.title = new Date(value).toString()` is a good idea. No matter using `timestampFormat` or not, we can keep this title.

src/js/Node.js
josdejong: Ah, that's a good point. You could probably solve it like 

```js
const title = (hasTimeStampFormat && timeStampFormat()) || defaultTimeStampFormat()
```
but it's ok with me either way :+1:

smallp: Yeah,  I have tried this solution, but I think that's too long and not so readable. Maybe that's because I'm a backend.:) Feel free to modify my code. You can modify it as your way.

josdejong:
Thanks again @smallp, I've merged your PR and will add the `title` again.

Issue #881, opened 2020-01-13
Changing sass variables


petermanders89:
I noticed that in the SCSS files the file styles is imported all over. This file contains variables, however it is not easy to overwrite them. Since the menu will still be blue:

```
$jse-blue: red;

@import "jsoneditor/src/scss/menu";
```

I am wondering why the variables within the styles.scss are not set with !default. With !default you can easily overwrite the styling in custom applications. Other libraries, like bootstrap for example, have most of their variables defined with the !default parameter. It would make overwriting the styling easier and friendly. The following example will still have a red color:

```
$jse-blue: red;
$jse-blue: blue !default;

@import "jsoneditor/src/scss/menu";
```

Not sure if I should re-open an separate issue for this, but if the !default parameter is added, I would suggest changing the name of the variable `$jse-blue` to something more generic.

josdejong:
Thanks for your suggestion. I'm all in for solutions who make it easier to customize the styling :)

I hadn't heard of `:default` before. Sounds good. The [page at MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/:default) suggests that Edge and IE don't support `:default`. Is this usable then? Or does SASS resolve this?

petermandersvin:
The MDN is referring to an actual css selector, which is a different default. SASS takes care of this. How it works in SASS, the !default will only activate if the variable is not already defined. SASS does take care of this. A simple example:

```
$jse-blue: red;
$jse-blue: blue !default;

#test{
  background-color: $jse-blue;
}
```

After compiling to CSS it will turn into:

```
#test {
  background-color: red;
}
```

josdejong:
Ah I understand it. Sounds good!

Anyone interested in implementing this improvement?

josdejong:
This is now available in `v8.2.0`. Dank je wel Peter :)

PR #882, opened 2020-01-15
Changing sass variables #881

petermanders89:
Pull request for implementing #881

### DIFF
--- a/src/scss/contextmenu.scss
+++ b/src/scss/contextmenu.scss
@@ -82,7 +82,7 @@
           height: 24px;
           padding: 0;
           margin: 0 4px 0 0;
-          background: url("./img/jsoneditor-icons.svg") 0 -72px;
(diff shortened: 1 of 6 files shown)


josdejong:
Thanks Peter! 

Small question: you've added `!default` to most of the variables but not all. Is there a specific reason not to add `!default` to `$jse-box-shadow...` and `$js2-font...` variables?

petermanders89:
Good question indeed!

I was aiming to focus on the colors, but indeed why not add it to all variables it to all variables. I also took the liberty to add an icons variable. I got carried away a bit while browsing threw the sass files. It might be that people want to customize the images or overwrite the path of the file considering there build system.

josdejong:
Nice idea to create a variable for the icons too! I think it's good that _everything_ is now consistently `!default`. Thanks!

josdejong:
Available now in `v8.2.0`

PR #885, opened 2020-01-17
Creating and optimized scss imports

petermanders89:
Opened this merge request to further improve the imports of SCSS. A couple of changes here:

- Created a root `jsoneditor.scss` which imports all the sub files. So in general, if you would need to import the entire styling of the editor you can import the one file, instead of all separate files 
- Moved all styling to a folder called jsoneditor. It is still possible to import single scss files
- Renamed `styles.scss` to `variables.scss`
- Updated the gulp file accordingly, which is simplified now. 

Upgrading to this will further improve the experience of users of scss. 

The `selectr.scss` is currently in the js folder, therefor mentioned seperatly in the gulp file. This is a bit unfriendly if people want to use the scss files. They would miss out on this styling. Is it an option to move this file also to the `scss/jsoneditor` folder? Perhaps in a subfolder? 

Note: No styling has changed or adjusted.


josdejong:
Thanks Peter, that's a neat improvement! 

The reason there is this folder `assets/selectr` is that the project was no longer maintained and had some issues. Ideally we should load the dependency via npm like all others. I think it's good to keep all selectr related stuff in this assets folder, so it's clear what to replace when getting it again via npm or migrating to a different library. 

I think what we can just do is import `selectr.scss` in `jsoneditor.scss`, so users don't have to bother. What do you think?

petermanders89:
That would indeed a nice solution. I added it. Tested it with the gulp builder and also in a separate webpack application.  Seems to be working fine.

josdejong:
Thanks, looks very good!

Two more thoughts:

1. It may be good to document somewhere that it's easy to customize styling with SCSS now . Maybe in the document [docs/styling.md](https://github.com/josdejong/jsoneditor/blob/develop/docs/styling.md)? (We can reorganize the docs if needed)
2. A question: what is the reason to start all scss file names with an underscore? Like `_editor.scss`? I'm not familiar with that naming pattern but it may be very common?

petermanders89:
1. I can look into that when I have a bit more time to create a simple document on how to overwrite styling etc. We could also consider renaming a couple of variables for there purpose. `$jse-blue` does not really say what the color is used for. It is the main color or primary color so that could be renamed to `$jse-primary-color` or `$jse-main-color`. Furthermore, I think in the documentation it should be clear that the `variables.scss` is required, allthough a bit obvious.  

2. Note that not all files are named with an underscore, only the once in the subfolder which are imported. There is a couple of reasons to do this. Mainly, as scss describes, it defines that the file is a partial. I think in this situation only the `_variables.scss` could be considered as a partial. Since its the only file a couple of other files are depending on.
The second reason has to do with compiling. If you are running sass compiler in the folder in which you would have `jsoneditor.scss` and a `main.scss`. By default you will get two css files; `jsoneditor.css` and `main.css`.If you name the file `_main.scss` it will be skipped. This does include subfolders. Files starting with an underscore are not compiled, by default. With gulp it is specified which file you want to compile, but this is not always the case.

josdejong:
1. Sounds good, giving the naming a bit more thought before documenting them and making them public :+1: 
2. Thanks, that makes sense, I like it.

josdejong:
@petermanders89 shall I merge this PR already, and we can look into documentation and renaming separately?

petermanders89:
@josdejong I think that would be good. I am sorry, I have been a bit busy passed couple of weeks. The focus of this pull request is only on enhancing the imports.

josdejong:
Ok thanks :+1: . Take it easy, I'm just very happy with the improvements you've made already :)

PR #927, opened 2020-03-09
 bugfix: error color and number color

reatang:


src/js/ace/theme-jsoneditor.js
```diff
@@ -102,7 +102,7 @@ color: #96DC5F
 color: darkorange
 }
 .ace-jsoneditor .ace_constant.ace_numeric {
-color: red
+color: rgb(0, 0, 205)
```
josdejong: Since all colors are using HEX code, it maybe good for consistency to also enter the color for numbers as HEX (`#ee422e`)

src/scss/jsoneditor/_variables.scss
```diff
@@ -5,7 +5,7 @@ $jse-blue: #3883fa !default;
 $jse-content-color: #1a1a1a !default;
 
 $jse-string: #006000 !default;
-$jse-number: #ee422e !default;
+$jse-number: #6897bb !default;
```
josdejong: What is the reason that you changed the colors for numbers from red to blue? (I would prefer to keep the red color)

josdejong:
Thanks for your PR @reatang .

Currently the variable `$jse-number` is used for coloring numbers as well as coloring error text. It's a good idea to separate this. I think though we should create a new `$jse-number` variable with the same color code as `$jse-number` (red), and rename the current `$jse-error` to `$jse-error-background`, since that's how it is used right now. Does that make sense?

src/scss/jsoneditor/_variables.scss
reatang: red just like Error 😂

src/js/ace/theme-jsoneditor.js
reatang: I borrowed the `ace editor` the color of the number.

reatang:
`jse-error-background` sounds like background-color rather than text-color.

josdejong:
> `jse-error-background` sounds like background-color rather than text-color.

That's correct: `$jse-number` was actually used only at one place for a background color, see _editor.scss. Therefore I think it makes sense to rename it. So:

1. rename 
    ```scss
    $jse-error: #ee2e2e70 !default;
    ```` 
    to 
    ```scss
    $jse-error-background: #ee2e2e70 !default;
    ````
2. create a new `$jse-error: #ee2e2e !default;`
3. Change the usages of `$jse-number` to `$jse-error` where applicable.

What do you think?

PR #956, opened 2020-04-13
docs: Fix simple typo, vaidation -> validation

timgates42:
There is a small typo in src/js/JSONEditor.js.

Should read `validation` rather than `vaidation`.


josdejong:
Thanks Tim 👍

PR #1098, opened 2020-08-29
Multi-select mouse events now honor the event's window

joshkel:
My app uses [window.open](https://developer.mozilla.org/en-US/docs/Web/API/Window/open) via [react-new-window](https://github.com/rmariuzzo/react-new-window) to show a JSONEditor instance in a separate browser window. When I interact with the editor and then close the separate window, I start getting JavaScript errors when I moved the mouse over the main window:

```
Uncaught TypeError: Cannot read property 'nodes' of null
    at v.H._onMultiSelectEnd (0:50503)
    at 0:50440
```

It appears that the problem is caused by JSONEditor's use of the `window` global variable - that causes any event handling to be done in the main window, so the child window doesn't get events and can't properly clean up. From reading MDN and doing some local testing, using [event.view](https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/view) instead seems to be the best fix.

I saw that the JSONEditor source code contains other references to the `window` global. I haven't seen problems from them yet, but I'm only using part of JSONEditor's functionality, and they may need to be fixed eventually to fully support use from `window.open`. (Using [Node.ownerDocument](https://developer.mozilla.org/en-US/docs/Web/API/Node/ownerDocument) and [Document.defaultView](https://developer.mozilla.org/en-US/docs/Web/API/Document/defaultView) from the editor's DOM node looks like it should work, although I haven't tested it.)

All of this seems like an unusual use case, so let me know if it's out of scope for JSONEditor, if you'd like a different approach, and/or if you'd like me to look into the other `window` uses.

Thank you.

### DIFF
--- a/examples/01_basic_usage.html
+++ b/examples/01_basic_usage.html
@@ -36,7 +36,8 @@
       'number': 123,
       'object': {'a': 'b', 'c': 'd'},
       'time': 1575599819000,
-      'string': 'Hello World'
+      'string': 'Hello World',
+      'onlineDemo': 'https://jsoneditoronline.org/'
     }
     editor.set(json)
   }
--- a/examples/24_new_window.html
+++ b/examples/24_new_window.html
@@ -30,12 +30,16 @@
       editor = undefined
     }
 
+    // make the necessary styles available within the child window
+    // for JSONEditor
     const baseUrl = window.location.href.slice(0, window.location.href.lastIndexOf('/'))
-
     const link = child.document.createElement("link")
     link.setAttribute("href", baseUrl + "/../dist/jsoneditor.css")
     link.setAttribute("rel", "stylesheet")
     child.document.head.append(link)
+    // for vanilla-picker
+    const colorPickerStyles = JSONEditor.VanillaPicker.StyleElement.cloneNode(true)
+    child.document.head.append(colorPickerStyles)
 
     const container = child.document.createElement("div")
     child.document.body.append(container)
--- a/examples/24_new_window.html
+++ b/examples/24_new_window.html
@@ -33,10 +33,10 @@
     // make the necessary styles available within the child window
     // for JSONEditor
     const baseUrl = window.location.href.slice(0, window.location.href.lastIndexOf('/'))
-    const link = child.document.createElement("link")
-    link.setAttribute("href", baseUrl + "/../dist/jsoneditor.css")
-    link.setAttribute("rel", "stylesheet")
-    child.document.head.append(link)
+    const jsonEditorStyles = child.document.createElement("link")
+    jsonEditorStyles.setAttribute("href", baseUrl + "/../dist/jsoneditor.css")
+    jsonEditorStyles.setAttribute("rel", "stylesheet")
+    child.document.head.append(jsonEditorStyles)
     // for vanilla-picker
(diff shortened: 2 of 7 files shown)


josdejong:
Interesting, I had never heard of `event.view` before. So apparently a window opened via `window.open` has a different global which is not `window`? Would be good to make the JSONEditor robust against this! 

I think it would be good to replace at least all mouse events which are using `window` with `event.view`. I'm not sure what we should do with cases like `window.getSelection()`, `window.innerHeight` `window.open`, and a few polyfills like for `window.Element`. Or is there simply an other "global" that we should use instead of `window`?

How can we easily test this? Do you have a minimal example demonstrating how JSONEditor opened via `window.open`?

josdejong:
@joshkel did you see my comment?

joshkel:
@josdejong I apologize for the delayed reply.

Here's a sandbox that demonstrates the problem:

https://codesandbox.io/s/jsoneditor-new-window-sus7d?file=/src/index.js

I'm having trouble reproducing the specific TypeError, but you can see some symptoms; the nodes of the tree can't be dragged and dropped, and if you mouse over the "sample text" text in the main browser window, you can see that it has a draggable cursor.

I'll see if there's a good fix for `window.getSelection()`, etc., although I may not have time until this weekend.

josdejong:
Thanks for working out the code example! I think it would be nice to add this example to the `/test` folder: there are more HTML examples there which make it easy to test some specific cases.

> I'll see if there's a good fix for `window.getSelection()`, etc., although I may not have time until this weekend.

Thanks! No need to rush

josdejong:
Nice job Josh 👍, thanks a lot. The example that you created is really helpful. I've done some extensive testing and all works like a charm. Thanks also for the additional errors and typos that you found and fixed, like with `showTransformModal` and calling `window.open`. 

I will merge your PR now.

josdejong:
I noticed that new window doesn't work on Internet Explorer. It would be great if you could have a short look to see whether this is fixable or not. If it would be a lot of work I'm not sure if we should try get it working.

josdejong:
Your fix has been published now in `v9.1.1` (works on all browsers except IE).

joshkel:
@josdejong Thanks for testing and releasing the fix. I apologize for not looping back with you on this.

I meant to test IE; thanks for letting me know. I did some investigating, and there are two main problems:

* I used `append` instead of `appendChild` in 24_new_window.html. That seems like a mistake, and it's easy to fix.
* IE apparently doesn't like it if you create a node (via `document.createElement` or similar) in one document object and append it to another document object; it throws a HierarchyRequestError if you try to do so. The fix would be to replace each of those calls with code that finds the appropriate document object of the JSONEditor instance. From looking briefly at the JSONEditor source code, that affects a _lot_ of code; I'm tempted to say that IE with new windows isn't supported, but maybe it's easier than it appears at first glance.

I'm happy to open a quick PR to replace `append` with `appendChild` and add a no-IE warning to 24_new_window.html if you want to go that route.

Thanks again.

josdejong:
Thanks for figuring this out. If we have to make `document.createElement` resolve the right `document` object dynamically, that would indeed be quite a refactoring. I agree with you, let's not put that effort into getting this working on IE. 

> I'm happy to open a quick PR to replace `append` with `appendChild` and add a no-IE warning to 24_new_window.html if you want to go that route.

Yes please, that would be very neat!

PR #1113, opened 2020-09-20
Update react advanced example

kalashnikovisme:
We should update text only once on the first render.
In other cases, the cursor will go to the beginning every time we pressing `Enter`.

### DIFF
--- a/examples/react_advanced_demo/src/JSONEditorReact.js
+++ b/examples/react_advanced_demo/src/JSONEditorReact.js
@@ -21,7 +21,7 @@ export default class JSONEditorReact extends Component {
       this.jsoneditor.set(this.props.json);
     }
   
-    if ((this.jsoneditor.getText() == "{}") && ('text' in this.props)) {
+    if ('text' in this.props) {
       this.jsoneditor.setText(this.props.text);
     }
     this.schema = cloneDeep(this.props.schema);
@@ -33,7 +33,7 @@ export default class JSONEditorReact extends Component {
       this.jsoneditor.update(this.props.json);
     }
 
-    if ('text' in this.props) {
+    if ((this.jsoneditor.getText() == "{}") && ('text' in this.props)) {
       this.jsoneditor.updateText(this.props.text);
     }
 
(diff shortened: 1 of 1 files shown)


josdejong:
Thanks for your PR Pavel! 

I'm not sure how pressing `Enter` moves the cursor to the beginning, how can I reproduce this issue? 

The fix you propose is in `componentDidMount`. This method is invoked only once when the component is created, so I'm not sure how this can cause an effect when pressing `Enter`. Can you explain?

kalashnikovisme:
@josdejong Oh! I'm sorry! Added condition to the wrong method. Of course, it should be at `componentDidUpdate`

Somehow pressing `Enter` in my case re-render the component, and as I understand after updating `text` editor moves the cursor to the beginning. I've added one more condition to the `componentDidUpdate`. It declines to update `text` if `text` already has been set.

Of course, we can update text from the state no more. I think we should add some comment about it.

**How to reproduce**

I just used your example without any changes with:
* React `16.9.0`
* Redux `4.0.5`

josdejong:
Thanks for the update. 

I think I'm still missing something: I updated the example `react_advanced_demo` to `react@16.9.0` (redux is not part of the example), started it, and tried to press enter in Tree mode, Code mode, and text mode, but the cursor does not jump to the beginning. How can I make that happen exactly?

kalashnikovisme:
I'm sorry. But I can't help you anymore :( I've tried to add your example to my project one more time without any change, but the result was the same. The cursor jumped to the beginning... :(

kalashnikovisme:
Maybe we can add some comments about troubleshooting to this example?
If it will happen again with somebody's project, it would be great if we'll help them.

josdejong:
Thanks for the update Pavel! 

So if I understand it correctly, the react example on it's own works correctly, but as soon as you copy the example in your application, the cursor jumps to the start when you press `Enter`. It may have to do with re-rendering or something, but without being able to reproduce it's hard to fix.

Let's close this PR until someone else also runs into this issue or you have time again for it.

kalashnikovisme:
I agree.

Hope, that nobody will face this issue :smiley:

Issue #1111, opened 2020-09-17
Enum dropdown not showing when using patternProperties for schema


ziga-miklic:
Hello,

I have the following schema:

```js
{
  title: 'Ports',
  description: 'Object containing ports',
  type: 'object',
  patternProperties: {
    '.*': {
      type: 'object',
      properties: {
        testrigId: {
          enum: this.devicePorts,
        },
      },
      required: [
        'testrigId',
      ],
    }
  },
};
```

The problem is that the `testrigId` property is not showing a dropdown, but a string input. A similar schema using `properties` instead of `patternProperties` is displaying the dropdown.

![issue](https://user-images.githubusercontent.com/3201207/93480738-0f484e80-f8fe-11ea-839f-7dfd3e47f555.png)
![working](https://user-images.githubusercontent.com/3201207/93480742-0fe0e500-f8fe-11ea-993d-353fe1545af8.png)

I see that this was fixed in #486, but I also see someone having the same issue this year in #904 but the issue was only closed.

Is this an issue with JSON Editor or am I doing something wrong?

Thank you!

josdejong:
Thanks for reporting Žiga. I just tried it out and I can reproduce this issue with the schema above, there is indeed no enum being displayed.

Anyone able to look into this issue? Help would be very welcome.

ziga-miklic:
Hello, @josdejong thank you for your response. I can give it a go and try to fix it.

I'll let you know how it goes.

josdejong:
Thanks 👍

josdejong:
Fix (#1116) has been published in `v9.1.1`, thanks again!

PR #1116, opened 2020-09-21
Fix for issue #1111

ziga-miklic:
Fixes enum dropdown not showing when using patternProperties for schema.

The issue happens as `_findSchema` is recursively called one more than than it should. Instead of returning the nested schema object (eg. `{ enum: [1, 2, 3] }`), it returns the parent object, eg.:

```javascript
{
  type: 'object',
  properties: {
    testrigId: {
      enum: [1, 2, 3]
    },
  },
  required: [
    'testrigId'
  ]
}
```

The fix adds a check to only call the `_findSchema` again if the object contains properties. 

The unit tests pass and I've also tested a complex schema inside `patternProperties` that uses the special types (boolean, color, enum).

### DIFF
--- a/src/js/Node.js
+++ b/src/js/Node.js
@@ -4448,7 +4448,7 @@ Node._findSchema = (schema, schemaRefs, path) => {
 
       if (typeof key === 'string' && childSchema.patternProperties && !(childSchema.properties && key in childSchema.properties)) {
         for (const prop in childSchema.patternProperties) {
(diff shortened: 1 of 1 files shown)


josdejong:
Thanks a lot Žiga, glad that you found the cause 👍

josdejong:
Fix has been published now in `v9.1.1`.

PR #1122, opened 2020-10-04
Fix unable to scroll list of keys

### LINKED ISSUE #1119: Unable to scroll the on the list of keys of the object.
Click on treepath breadcrumb, 

image

List of all the keys in the object shows up, from where we can navigate to a particular key of that JSON, but unable to scroll when there are more keys:

The following screenshot shows object having around 50 keys but only 22 items are accessible and others are hidden.

image

tanmayrajani:
- fixes #1119

### DIFF
--- a/src/js/ContextMenu.js
+++ b/src/js/ContextMenu.js
@@ -9,8 +9,10 @@ import { translate } from './i18n'
  * @param {Object[]} items    Array containing the menu structure
  *                            TODO: describe structure
  * @param {Object} [options]  Object with options. Available options:
- *                            {function} close    Callback called when the
- *                                                context menu is being closed.
+ *                            {function} close        Callback called when the
+ *                                                    context menu is being closed.
+ *                            {boolean} mainMenuBar   Whether main menu bar is
+ *                                                    enabled or not.
  * @constructor
  */
 export class ContextMenu {
@@ -24,6 +26,7 @@ export class ContextMenu {
     this.eventListeners = {}
     this.selection = undefined // holds the selection before the menu was opened
     this.onClose = options ? options.close : undefined
+    this.mainMenuBar = options ? options.mainMenuBar : false
 
     // create root element
     const root = document.createElement('div')
@@ -243,7 +246,8 @@ export class ContextMenu {
     }
 
     if (anchor.className.includes('jsoneditor-treepath-seperator')) {
-      const maxPossibleMenuHeight = frameRect.height - parentRect.top - parentRect.height
+      const mainMenuBarHeight = this.mainMenuBar ? 35 : 0
+      const maxPossibleMenuHeight = frameRect.height - mainMenuBarHeight - parentRect.height
       this.dom.list.style.maxHeight = maxPossibleMenuHeight + 'px'
       this.dom.list.style.overflowY = 'auto'
       this.dom.list.style.width = '100%'
--- a/src/js/ContextMenu.js
+++ b/src/js/ContextMenu.js
@@ -11,8 +11,8 @@ import { translate } from './i18n'
  * @param {Object} [options]  Object with options. Available options:
  *                            {function} close        Callback called when the
  *                                                    context menu is being closed.
- *                            {boolean} mainMenuBar   Whether main menu bar is
- *                                                    enabled or not.
+ *                            {boolean} limitHeight   Whether ContextMenu height should be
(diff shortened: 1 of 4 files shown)


josdejong:
Thanks Tanmay!

The fix works like a charm. I noticed two small things which would be great to improve:

1. The scrollbar is rendered on top of the text of the menu items
2. It looks like the left over space at the bottom, below the menu, is always around 100px. Maybe we can make that smaller, like 10px or 20px or so, so that we utilize the space better in case of a small editor?

afbeelding

josdejong:
Ping @tanmayrajani do you have time to look into my feedback? Would be great

tanmayrajani:
Hi @josdejong, sorry, I totally forgot about this. I fixed the width issue and applied it to the treepath seperator only (as contextmenu may be used for other things as well I guess?)

re:
> It looks like the left over space at the bottom, below the menu, is always around 100px. Maybe we can make that smaller, like 10px or 20px or so, so that we utilize the space better in case of a small editor?

I couldn't re-create this issue. I tried two ways (pre-fill the JSON object on load, and, add key-value pairs one-by-one to increase the height), it worked in both of those ways for me.
image

josdejong:
Thanks for the update Tanmay.

FYI: the context menu is currently used in one other place: the menu to switch mode.

1. It looks like the scrollbar is still overlapping text but only on Firefox, not in Chrome (not sure if it was happening before in Chrome or not).
2. What I did was checkout your branch, run `npm run build`, and then open the file `test/test_build.html`. On both Chrome and Firefox, the menu stops roughly 100px above the bottom of the editor. The effect is even easier to see when you change the height of the editor from 500px to say 300px.

tanmayrajani:
Thanks for the details! I think I understood and fixed both the problems properly this time. :crossed_fingers:

josdejong:
Thanks for the update! It's now working indeed, I think we're getting there :)

I think we can do a simplification step. 

Inside the context menu there should be no need to have knowledge about something like `jsoneditor-treepath-seperator`. Based on the positions and size of the `frame` and `anchor`, we can calculate how much space there is:

- When the context menu is displayed below the anchor, the available space is simply: 

    ```js
    const margin = 10 // make sure there is a little margin left
    const maxPossibleMenuHeight = frameRect.bottom - anchorRect.bottom - margin
    ```

- When the context is displayed on top of the anchor, the available space is:
    ```js
    const margin = 10 // make sure there is a little margin left
    const maxPossibleMenuHeight = anchorRect.top - frameRect.top - margin
    ```

I noticed that the line `this.dom.list.style.width = '100%'` works for Chrome but not for Firefox, and when applied to the ContextMenu shown when clicking the buttons left from any of the nodes, it even looks bad. Maybe we can simply introduce a boolean flag `limitHeight: true | false` for ContextMenu, have this `false` by default, and only set this to `true` for the context menu of the treepath. What do you think?

tanmayrajani:
Thank you for the suggestions, makes much more sense, I should've thought of this on my own though :disappointed: 

Also, is it a valid scenario for contextmenu on the treepath to render upwards (top of the anchor)? I mean, you have mentioned it so it would most likely be.. but I'm not sure how I could replicate it to test..

josdejong:
> Also, is it a valid scenario for contextmenu on the treepath to render upwards (top of the anchor)? I mean, you have mentioned it so it would most likely be.. but I'm not sure how I could replicate it to test..

The navigationBar (tree path) is always displayed on top, so the menu should always be rendered below. So I think we should not worry about that now, until some day in the future we allow rendering the navigationBar at the bottom of the editor but that's just theory at this point :).

tanmayrajani:
I think I fixed it the way you suggested. I checked in firefox too, looks ok to me so far. Could you have a nother go at it and let me know? Thank you!

josdejong:
Thanks for the update @tanmayrajani , that looks simple and straightforward 👍 . Works like a charm indeed.

One last small thing: is it possible to remove the following line:

```js
this.dom.list.style.width = '100%'
```

It looks like it causes a small white offset right from the selected entry with red background: 

afbeelding

Or would that break something else?

tanmayrajani:
sorry for a big delay! I checked a few things, it didn't break anything else.

josdejong:
Thanks Tanmay, no problem at all! Will merge your fix now 🎉

josdejong:
Published now in `v9.1.4`

Issue #1158, opened 2020-11-04
Error: foundSchema is null


maufl:
Hey, it looks like I was able to trigger a bug in `Node._findSchema`. I open a JSON file in the editor configured with a schema I provide. When I open child nodes in the JSON document, an error is thrown inside of jsoneditor. The error is `Error: foundSchema is null` in `Node.js:4451`.

This is a reduced version of the schema I'm using:
```json
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "type": "object",
    "patternProperties": {
        "^/[a-z0-9]*$": {
            "$ref": "#/definitions/component"
        }
    },
    "definitions": {
            "component": {
                "type": "object",
                "properties": {
                    "type": {
                        "type": "string",
                        "minLength": 1
                    },
                    "config": {
                        "type": "object"
                    },
                    "children": {
                        "type": "object",
                        "patternProperties": {
                            "^/[a-z0-9]+$": {
                                "$ref": "#/definitions/component"
                            }
                        }
                    }
                }
            }
    }
}
```

This is a reduced version of the JSON document I'm loading:
```json
{
    "/": {
        "type": "MessageOfTheDay"
    },
    "/status": {
        "type": "TabLayout",
        "children": {
            "/status":{
                "type": "StatusComponent",
                "config": {
                    "fields": []
                }
            },
            "/bus":{
                "type": "StatusComponent",
                "config": {
                    "fields": []
                }
            },
            "/system":{
                "type": "StatusComponent",
                "config": {
                    "fields": []
                }
            }
        }
    }
}
```

I tried to debug it but I'm so far unable to find the root error case. What I found out so far, is that JSON editor seems to not handle `$ref`s that point to a part of the same schema. This seems to be the reason why this error occurs when I expand the `children` node. I can work around it by explicitly  setting `schemaRefs: { "#/definitions/component": schema.definitions.component }` in the editor options. However I still get the error when expanding the `config` field.

One thing I noticed, is that `_findSchema` does iterate over the parts of the path to the node **and also** recursively call itself with sub paths to the node. This is also the reason why I don't really understand how `_findSchema` is supposed to work.

josdejong:
Thanks for reporting and debugging this Felix. 

Sounds indeed like a bug in `Node._findSchema`. Anyone able to dig further to figure this out? Help would be welcome.

maufl:
[...]

PR #1161, opened 2020-11-10
Fix schema not found if using internal references

maufl:
This merge request fixes issue #1158 
I refactored the `Node._findSchema` method, which now should correctly resolve internal references. I also added a test case for this.

### DIFF
--- a/src/js/Node.js
+++ b/src/js/Node.js
@@ -4451,56 +4451,54 @@ Node._findSchema = (topLevelSchema, schemaRefs, path, currentSchema = topLevelSc
   const nextPath = path.slice(1, path.length)
   const nextKey = path[0]
 
-  let possibleSchemas = currentSchema.oneOf || currentSchema.anyOf || currentSchema.allOf || [currentSchema];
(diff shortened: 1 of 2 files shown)


josdejong:
Thanks a lot for this fix Felix. Your refactoring looks very neat. And I like it to see more `const` 😉 .

One question: in the old version there was an iteration:
```js
for (let i = 0; i < path.length && childSchema; i++) {
  // ...
}
```
This iteration seems gone now, did you replace it by a recursive solution or something? Or was it redundant code?

maufl:
> This iteration seems gone now, did you replace it by a recursive solution or something? Or was it redundant code?

Yes, the old code used both an iteration with a for loop and one with a recursive call. I don't know why, to me it seemed to be redundant. The new code only does the recursive iteration. As it did not break any tests, I assume that it really was unnecessary duplication.

josdejong:
Wow, that is a very nice simplification then. This function is one of the few things that is well tested in the library, so all should be good indeed. I've also checked whether the JSON Schema examples still work, and all looks good to go 👍 .

josdejong:
Fix is now published in `v9.1.3`, thanks again Felix

PR #1239, opened 2021-02-09
Added support for references to another scheme

Hagartinger:
According to Json-schema documentation: 

> $ref can resolve to a URI that references another file, so if you prefer to include your definitions in separate files, you can also do that.

Extended _findSchema function's logic to work in such case.

### DIFF
--- a/src/js/Node.js
+++ b/src/js/Node.js
@@ -4472,6 +4472,15 @@ Node._findSchema = (topLevelSchema, schemaRefs, path, currentSchema = topLevelSc
             throw Error(`Unable to resovle reference ${ref}`)
           }
         }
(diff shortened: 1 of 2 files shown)


josdejong:
Thanks @Hagartinger , that looks like a nice improvement!

Can you please fix the linting issues (`npm run lint` and `npm run lint -- --fix`), and can you write a unit test for this new functionality in the file `/test/Node.test.js`?

Hagartinger:
Here you go. Feel free to tell me if you want to have anything changed :-)

josdejong:
Thanks @Hagartinger , looks good 👍

josdejong:
Your fix is now published in `v9.1.10`

PR #1244, opened 2021-02-15
Add ajv-i18n support

askalione:
Hello! Add i18n for validation error messges using ajv-i18n.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1404,6 +1404,11 @@
         "uri-js": "^4.2.2"
       }
     },
+    "ajv-i18n": {
+      "version": "3.6.0",
+      "resolved": "https://registry.npmjs.org/ajv-i18n/-/ajv-i18n-3.6.0.tgz",
+      "integrity": "sha512-F21DzmzYq9aVtY8CGGtlnQDy3rFMaFW2KRlMuCQp76KiPIkvqN+mpq5MI9EsgC0VFwj+jeLIsvVCvPOa1sobBQ=="
+    },
     "ajv-keywords": {
       "version": "3.5.2",
       "resolved": "https://registry.npmjs.org/ajv-keywords/-/ajv-keywords-3.5.2.tgz",
--- a/package.json
+++ b/package.json
@@ -28,10 +28,11 @@
   "dependencies": {
     "ace-builds": "^1.4.12",
     "ajv": "^6.12.6",
+    "ajv-i18n": "^3.6.0",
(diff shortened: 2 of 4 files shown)


josdejong:
Thanks for your PR Alexey, this will be a nice improvement.

The PR itself is very hard to see what actually changed: about every line in `textmode.js` and `treemode.js` is changed. Can you please maintain the 2 space indentation, and if you would like to do any refactorings in the code do that in a separate PR?

josdejong:
Please note that you can run `npm run lint` to detect linting issues, and use `npm run lint -- --fix` to fix as much as possible automatically.

askalione:
> and use npm run lint -- --fix to fix as much as possible automatically.

Thanks!

I have fixed indents issues and now all checks have passed. Should i make new PR?

josdejong:
Thanks for the update, the code looks good 👍 . No need to create a new PR, I'll squash it when merging. 

I'll test your work functionally soon.

josdejong:
I've done some testing, and it works like a charm, thanks @askalione!

Two feedbacks:

1. Bundling the full `ajv-i18n` is quite large: the minified+zipped bundle goes from 231KB to 244KB. I'm wondering whether we can come up with a solution which doesn't require bundling of all the locales. The only idea I can come up with is to have the user pass the right translation file himself via the options, like:

    ```
    const options = {
      language: 'ru'
      ajv-i18n: require('ajv-i18n/localize/ru')
    }
    ```

    But that is not easy to use. Do you have any ideas in this regard? If not I guess we simply have to accept this for now.

2. JSONEditor has a minimalist version of the bundle which basically strips the Ace and Ajv, and color picker libraries. Since those are quite large. The `ajv-i18n` library is also stripped, but now the `dist/jsoneditor-minimalist.js` throws an error when you open the editor with a JSONSchema. We can solve this by putting a try/catch around the `require` loading `ajv-i18n`, the same way as is done for Ajv itself in the file `tryRequireAjv.js`. Can you adjust that and make sure the code doesn't break when `ajv-i18n` is not available in the bundle?

johanbm1405-commits (approved):

josdejong:
@johanbm1405-commits why do you suddenly "Approve" all open PR's?

johanbm1405-commits (approved):

johanbm1405-commits (approved):

josdejong:
@johanbm1405-commits please stop marking these PR's as "approved".

PR #1252, opened 2021-02-24
Add Dockerfile

rgarrigue:
Hi @josdejong 

I like JSONEditor, and since I like self hosting my tools too, here's a Dockerfile I made. It works like this

```bash
docker build -t local/jsoneditor:test .
docker run --name test -p80:80 -d local/jsoneditor:test
# open localhost:80 in your browser
docker rm -f test
```

I guess this could be added somewhere in the documentation.

The `misc/index.html` is a c/c of the README's example with a <title>. The result is bad, I hope you can provide a better one, a bit like the official jsoneditor website ?

Last but not least, I can provide a github action to build & push this into docker hub for an official image. Except you'ld need and official repository, getting one for FOSS project is quite easy. So if you're interested, tell me, I'll add it to the PR

Fare well :-)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,11 @@
+FROM node:current-alpine AS builder
+WORKDIR /opt
+COPY package*.json /opt/
+RUN apk add python make g++ && \
+    npm install
+COPY . /opt/
+RUN npm run build
+
+FROM nginx:alpine
+COPY --from=builder /opt/misc/index.html /opt/dist/jsoneditor.min.js /opt/dist/jsoneditor.min.css /opt/dist/img/jsoneditor-icons.svg /usr/share/nginx/html/
+EXPOSE 80
--- a/misc/index.html
+++ b/misc/index.html
@@ -1,0 +1,33 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <title>JSONEditor</title>
+    <link href="jsoneditor.min.css" rel="stylesheet" type="text/css" />
+    <script src="jsoneditor.min.js"></script>
+  </head>
+  <body>
+    <div id="jsoneditor" style="width: 400px; height: 400px"></div>
+    <noscript> You need to enable JavaScript to run this app. </noscript>
+    <script>
+      // create the editor
+      const container = document.getElementById("jsoneditor");
+      const options = {};
+      const editor = new JSONEditor(container, options);
+
+      // set json
+      const initialJson = {
+        Array: [1, 2, 3],
+        Boolean: true,
+        Null: null,
+        Number: 123,
+        Object: { a: "b", c: "d" },
+        String: "Hello World",
+      };
+      editor.set(initialJson);
+
+      // get json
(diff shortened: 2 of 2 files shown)


josdejong:
This is an interesting idea Rémy, thanks! 

It would be more neat indeed to give the editor a full page size instead of 400x400 px (that's easy to adjust).

Thinking aloud here: as long as this is just a single Dockerfile and index.html file it could be part of this repo, but as soon as you want to do do steps like push an image to docker hub etc, it is probably better to do that in a separate repo so both repo's can focus on one thing. So we could set up a repo `jsoneditor-docker` or `docker-jsoneditor` or something like that and take it from there. What do you think?

josdejong:
Why would it be easier from this repo `jsoneditor` vs a separate repo like `jsoneditor-docker`? I suppose in both cases it is a single `npm install ...` and then a docker build?

rgarrigue:
Let's say you put the Dockerfile in a separate repository. If the CI is there too like Github Action or Docker Hub integratin, how do you know in this context when a tag is made here ? You don't. So what, a cron ? Costly if possible. I'ld say it's an enterprise solution, when you have a generic way of doing thing with a standard Dockerfile for most of your react app, provided they follow the enterprise good practice, etc.

Solution n°2, you put the CI definition hence triggering here. Half baked since you've the CI part of the docker thing here, and the rest in the separate repository. But it's able to trigger on a tag event basis, good. But what will you do then, checkout the separate repository to grab your dockerfile & assets. More scripting just to separate the Dockerfile & asset ? Drifting far from KISS here for no good reason.

Again, it's just my opinion. Your app, your choice.

josdejong:
Ah, so you're thinking about how to automate publishing a new version to Docker Hub when there is a new version of mathjs, I get it.

I'm not really sure, but it _may_ be not that hard to do in a separate repo: To keep a dependency up to date, you can use for example https://dependabot.com/. This automatically creates a PR when there is a new version of a dependency, which you can merge with one click. I can imagine that after merging, you can automatically run a Github Action which will publish a new version to Docker Hub.

The reason I'm a bit hesitant to add docker related stuff to the repository boils down to the single-responsibility principle. I suppose the integration to publish to Docker Hub requires adding some dependencies to the library, which bloat the library. And this year it's Docker, next year it's Kubernetes, or platform X and environment Y.

At this point the Docker Hub integration feels quite niche, though I understand it can be very nice to have. I have a preference to put the Docker integration in a separate repo. If Docker turns out to be a major solution for a lot of people, we can always reconsider integrating it in the `josdejong/jsoneditor` repo if that would be helpful.

rgarrigue:
Hi @josdejong. I'm cleaning up my stale PRs list. What about this one ? If you decide to add it here it'll need fixes on the index.html, otherwise you can close it and c/c the Dockerfile to a docker dedicated one whenever you feel like

josdejong:
Thanks for getting back Rémy.

I prefer creating a separate repository like say `jsoneditor-docker` for this to keep the current repo focused. Thanks again for your idea and PR though! I'll close this PR now.

PR #1268, opened 2021-03-12
css URL update to be into variables

ppetkow:
Commits on Mar 11, 2021 needs to be cherrypicked 
There is a major problem with create react app if someone tries to override background images in CSS. 
The solution is to move URL() into the variable in scss. 
This commits fix critical issues with the css-loaders package into react-scripts.


josdejong:
Thanks for your PR @ppetkow 👍 

I see there is almost 1000 new lines of styling, it looks like all styling is duplicated now. All styling was in files with a prefix `_`, and now there are new files without this prefix. Shouldn't these changes be made in the existing files with `_` prefix?

ppetkow:
maybe is better to make a new fork .. only the last 3 commits and that's the correct changes

ppetkow:
I clear scss files and update  autocomplete  files with colors .. i hope this is ok now

josdejong:
That looks good, thanks!

Right now this is a breaking change, right?  If you where overriding `$jse-icons` before, your styling will be broken now. I was thinking whether it is possible to keep this backward compatible. Maybe we could do this:

```scss
$jse-icons: "./img/jsoneditor-icons.svg" !default;
$jse-icons-url: url($jse-icons) !default;
```

and use `$jse-icons-url` everywhere in the styling. Would that work?

ppetkow:
yep this will be the solution

josdejong:
Thanks @ppetkow , looks good 👍

josdejong:
Published now in `v9.3.1`. Thanks again!

PR #1300, opened 2021-04-12
Added Korean language support

luasenvy:
add i18n Korean Language Support :)


josdejong:
Thanks @luasenvy 👍

josdejong:
Korean language support is now available in `v9.4.0`

Issue #1072, opened 2020-08-05
How can I get the real-time validation status?


liyiming22:
How can I get the real-time validation status? (Error or No Errors)
For example: I set a button beside the jsoneditor, and when users input the wrong format json data, the button should be disabled, and when users fix the format issue, the button should be enabled.

I have seen the API documents, but I didn't find the apis like (getValidationStatus or status eg.)

josdejong:
Is `onValidationError` what you mean? It's called when there is a change in the errors (also from an error to no error)

Docs: https://github.com/josdejong/jsoneditor/blob/develop/docs/api.md

ChrisAcrobat:
For me, no. I don't think so.
I use `onChange` to listening for changes, and some parts can't be executed without knowing the current status. Including schema, so a simple `try{JSON.parse}` would not be sufficient.
It would be possible if it is promised that `onValidationError` is executed before `onChange`, but I can't find any confirmation abort that in the documentation.
So a `{function} getValidationErrors(errors)` would be handy.

Just for reference, in Ace you can get the information on any change with this:
``` JavaScript
aceEditor.getSession().on('changeAnnotation', ()=>{
	console.log(_editor.getSession().getAnnotations().filter(a=>a.type==='error'));
});
```

ChrisAcrobat:
After a quick look around, I think it would be sufficient to do `function getValidationErrors(){return this.lastSchemaErrors}` in `textmode.js` and `treemode.js`. I will look in to it more next time I have time.

josdejong:
👍

ChrisAcrobat:
After a bit more digging (way to much, because I missed the obvious 😅) I found that `this.lastSchemaErrors` is already available with `jsonEditor.lastSchemaErrors`. 🤦‍♂️
But as I feared, it seams to always be updated after `onChange` has been called.
The investigation continues. 🔎

josdejong:
Just an other idea: you can always create your own instance of Ajc with the same JSON schema and execute that at the moment you want with the current JSON.

ChrisAcrobat:
Yeah, I thought of that as well. I but at the same time, I thought that it should already be include somehow and did't want to reinvent the wheel.
I think I have found the problem. `onChange` is always executed before, because the validation is denounced. Returning a promise from the denounce should do the trick.

ChrisAcrobat:
[...]

PR #1313, opened 2021-05-08
Made onChange execute after debounced validation and added function to get valication errors.

ChrisAcrobat:
Added #1072

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -58,6 +58,10 @@ Constructs a new JSONEditor.
   > results in JSONEditor breaking because the format of the Ajv errors differs
   > from what is expected.
 
+- `{Function} getValidationErrors ()`
+
+    This function returns a promise that will resolve a array with the latest validation errors.
+
 - `{function} onChange()`
 
   Set a callback function triggered when the contents of the JSONEditor change.
--- a/src/js/JSONEditor.js
+++ b/src/js/JSONEditor.js
@@ -266,7 +266,7 @@ JSONEditor.prototype.getName = function () {
  * Get last errors.
  */
 JSONEditor.prototype.getValidationErrors = function () {
-  return this.validationErrors_promise;
+  return this.validationErrors_promise
 }
 
 /**
--- a/src/js/JSONEditor.js
+++ b/src/js/JSONEditor.js
@@ -263,6 +263,13 @@ JSONEditor.prototype.getName = function () {
 }
 
 /**
+ * Get last errors.
+ */
+JSONEditor.prototype.getValidationErrors = function () {
+  return this.validationErrors_promise;
+}
+
+/**
  * Change the mode of the editor.
  * JSONEditor will be extended with all methods needed for the chosen mode.
  * @param {String} mode     Available modes: 'tree' (default), 'view', 'form',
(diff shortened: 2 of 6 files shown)


josdejong:
Thanks for your PR Christoffer!

I think this PR contains two ideas:

1. Implement a method `getValidationErrors`.
2. Change `onChange` such that it triggers after the debounced JSON schema validation did run. 

I like the new method `getValidationErrors`, that makes a lot of sense. I have doubts about the changes in `onChange`. The callback `onChange` fires when the JSON contents changes. This is independent of changes in the JSON Schema validation, which can be captured with `onValidationError`.  Running JSON Schema validation is run debounced for performance reasons, so there is no direct coupling between the two.

If I understand your use case correctly, When `onChange` is triggered, you want to know if there are any validation errors and depending on that do something. I think the new function `getValidationErrors` that you create is perfect for this, so you will do something like:

```js
const options: {
  onChange: () => {
    const errors = editor.getValidationErrors()
    if (errors.length === 0) {
      const json = editor.get()
      // do something, like persisting the JSON
    }
  }
}
```

I think there is no need to wait with firing `onChange` until the debounce validation is executed (just keep the old behavior), and instead we can simply trigger not-debounded validation inside `getValidationErrors` if needed, something like:

```js
JSONEditor.prototype.getValidationErrors = function () {
  // run validation directly if still pending
  this._debouncedValidate.flush()

  // return the (up to date) validation errors
  return this.lastSchemaErrors
}
```

What do you think?

ChrisAcrobat:
I can try it, but I think that when I prototype it that the promise that I got back was the previous `getValidationErrors`-promise that was not up to date. But since it was just a prototype it might have ben something else that was wrong.
I'll reply after I have taken a look.

ChrisAcrobat:
I have reverted the changes of 2fa1b50b8e80a7345c296b8920cd1870b93fab88 and 7f006071e3468b06236e66fef218f1de5b7874a3 and, I think, solved the the same issue by storing the promise's resolve callback instead of the promise. But I would like to test it some more before I suggest it to be merged. I'll reply as soon as I have got the time.

josdejong:
@ChrisAcrobat this is a bit more complex than I expected, I forgot that validation can be async, and that the uses `debounce` function is a custom made and not the Lodash one which has this nice `flush()` method.

I also tried to work out a solution, see #1319. Basically I would like to have the existing `.validate()` method return a promise resolving with the errors. What do you think?

ChrisAcrobat:
I haven't had time to fully test my solution yet, but when I do I can test both and say how they work.
But in short, yes it should work if it also is debounced correctly.
I also realised that maybe my solution with just storing the call back maybe isn't that optimal. In theory it might happen that two callbacks executes in the wrong order, making the second one (first to execute) reporting an error and the first (later to execute) reporting no error, resulting in a report that says the second one with error contains no error.

josdejong:
Yes indeed, that makes it complicated with this combination of debouncing _and_ async. Therefore I thought I may be easier to simply "force" running the validation again, independent of the debouncing functionality, to keep things simple and reliable.

josdejong:
Closing this PR in favor of #1319. Thanks again!

PR #1319, opened 2021-05-12
Implement .validate() returning a promise

josdejong:
A second approach to solve #1313, #1072

Basically: let the existing `.validate()` method return a promise resolving the errors (if any).

Usage example:

```js
const options = {
  onChange: () => {
    editor.validate().then(errors => {
      if (errors.length === 0) {
        const json = editor.get()
        // do something, like persisting the JSON
      } else {
        // show error to the user or something
        console.log('validation errors', errors)
      }
    })
  }
}
```

Todo: 

- Thorough testing
- Document method `validate`

### DIFF
--- a/src/js/textmode.js
+++ b/src/js/textmode.js
@@ -97,7 +97,7 @@ textmode.create = function (container, options = {}) {
   this.lastSchemaErrors = undefined
 
   // create a debounced validate function
-  this._debouncedValidate = debounce(this.validate.bind(this), this.DEBOUNCE_INTERVAL)
+  this._debouncedValidate = debounce(this._validateAndCatch.bind(this), this.DEBOUNCE_INTERVAL)
 
   this.width = container.clientWidth
   this.height = container.clientHeight
@@ -324,7 +324,7 @@ textmode.create = function (container, options = {}) {
   this.errorTable = new ErrorTable({
     errorTableVisible: this.mode === 'text',
     onToggleVisibility: function () {
-      me.validate()
+      me._validateAndCatch()
     },
     onFocusLine: function (line) {
       me.isFocused = true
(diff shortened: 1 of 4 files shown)


ChrisAcrobat:
Hello again!
Just tested it, in short it seams to work fine, but it currently misses if it first is a valid Json that matches onValidate , and then turn it invalid (e.g. add a character outside of `{}`)
- First: Return no error.
```Json
{
  "number": 1
}
```
- Then: Still return no error.
```Json
{
  "number": 1
}a
```

josdejong:
Ahh, good find Chris! 

This should be fixed now in commit https://github.com/josdejong/jsoneditor/pull/1319/commits/2dd34aad5b79ce2754c5bd7ebdd9f3dfdfa375c3: changed the code to trigger `onValidationError` and then update `this.lastSchemaErrors` to match the code in `validateCustom(...).then(...)` (it was only working when `onValidationError` was defined, but should work regardless of that)

josdejong:
What do you think @ChrisAcrobat, is this PR a good solution for you, or do you think it's better to work out your PR #1313 further?

ChrisAcrobat:
I tried it now and found a problem. When i paste a full Json with a wrong ending character (like before) it misses the catch and reports as validated, but when fails later on `editor.get()`.
Removing the wrong ending character and insert it by it self is cached, just not when it replace the current with a paste, for some reason.
I think it is because [this.getText()](https://github.com/josdejong/jsoneditor/blob/2dd34aad5b79ce2754c5bd7ebdd9f3dfdfa375c3/src/js/textmode.js#L884) is first returning `""` and counts as a false and sidestepping assigning of `parseErrors`.
But `parseErrors` is later assigned, because `textmode.validate` is executed tree times when I pasted the invalid Json, so I guess there is some kind of race condition involved.
This does not happen with #1321, but I am not sure why (I have not have time to look in to it) because they both should suffer the same race condition.

ChrisAcrobat:
Have their been any progress? I haven't had time to look at this for a week. Maybe tomorrow.

josdejong:
> I haven't had time to look at this for a week.

Same here 😂 . Looking into it now.

josdejong:
@ChrisAcrobat I've been trying to reproduce the issue you describe in https://github.com/josdejong/jsoneditor/pull/1319#issuecomment-841831371, but no luck so far. Can you more exactly describe how to reproduce the issue you encountered?

I've updated [`test/test_schema.html`](https://github.com/josdejong/jsoneditor/blob/51b243558e5e8f8c696f3595dbbb7e4a0546c239/test/test_schema.html) to use the new `editor.validate()`, this example is what I've been testing with, copy/pasting valid and invalid JSON. 

It is indeed that case that when the editor is empty, no validation is done: it would be quite intimidating to see big red parse errors every time you open a new, empty document. But as soon as I paste some (invalid) JSON, `onChange` triggers again and I can call `const errors = await editor.validate()` again, returning the expected parse error.

ChrisAcrobat:
Sorry for the late reply. I will see if I can repoduce the error again and see if I can make https://github.com/josdejong/jsoneditor/blob/51b243558e5e8f8c696f3595dbbb7e4a0546c239/test/test_schema.html catch it.
I have for the moment forgotten what the problem was, but I will reply as soon as I have an update.

ChrisAcrobat:
@josdejong I can't neither reproduce it. Maybe I was running old code. If it occurs again I'll report it again then.
Thanks for the feature!
Just as a side note: It looks like [cdnjs.com](https://cdnjs.com/libraries/jsoneditor) has not been updated to 9.4.2.

josdejong:
Thanks for testing this thoroughly Chris and for thinking and working along on this!

I will merge this PR now and close #1313 in favor of this one.

josdejong:
Published now in `v9.5.0`

ChrisAcrobat:
No problem! Happy to help!

PR #1333, opened 2021-05-27
Rename spin animation

mdix:
The spin animation defined by selectr collides with the spin
animation defined by tailwind. It overrides the animation and causes
the spinning elements to spin offcenter. By renaming it we kinda
namespace it and prevent a collision.

[Definition of tailwind spin animation](https://tailwindcss.com/docs/animation)

### DIFF
--- a/src/js/assets/selectr/selectr.scss
+++ b/src/js/assets/selectr/selectr.scss
@@ -405,16 +405,16 @@
     -webkit-transform-origin: 50% 0 0;
     transform-origin: 50% 0 0;
 
-    -moz-animation: 500ms linear 0s normal forwards infinite running spin;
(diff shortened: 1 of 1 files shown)


josdejong:
Thanks, makes sense. 

Maybe a good idea if tailwind would also use prefixes for their class names 😁 . I remember Bootstrap also giving these kind of conflicts due to too generic class names.

mdix:
Hey! Thanks for merging. Actually, they did for v2 (https://github.com/tailwindlabs/tailwindcss/pull/2621), but v1.9.6 is unprefixed and if you need IE11 support you can't update.

josdejong:
Ahh that's good to hear!

josdejong:
Fixed now in `v9.4.2`

PR #1365, opened 2021-08-25
Use noreferrer for window.open

rajitbanerjee:
#### Notes
- Using `noopener` without `noreferrer` is a security risk with a threat of Reverse Tabnabbing. 
- Changing the flag to `noreferrer` which [automatically sets](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#window_features) `noopener` as well. 

#### Testing
- No linting issues and all tests pass.

### DIFF
--- a/src/js/Node.js
+++ b/src/js/Node.js
@@ -2581,7 +2581,7 @@ export class Node {
             // if read-only, we use the regular click behavior of an anchor
             if (isUrl(this.value)) {
               event.preventDefault()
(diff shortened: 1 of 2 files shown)


josdejong:
Thanks Rajit 👍 

I didn't know `noreferrer` automatically sets `noopener`, that is nice and clean

josdejong:
Fix published in `v9.5.4`

josdejong:
thanks again 😄

PR #1369, opened 2021-08-27
Add support for external validation

chaitu-tk:
In some usecases that need control on when a validation is triggered
there is no way to render the resutls of that validation and custom
validation handler cannot be used as that will be called for every
change. E.g., Save/Cancel based validation outside the editor.

Add two new APIs (set/clear)ValidatonErrors to update the results of the
external validation, these errors will be merged with other validation
errors and rendered all together.

Add an example to demonstrate the usecase and document the APIs.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -210,7 +210,20 @@ Constructs a new JSONEditor.
     }
   }
   ```
-  
+
+- `{function} setValidationErrors(errors: ValidationError[])`
+  Set the validation errors from external sources. Available in all modes.
+
+  Typically the editor is embedded into an external site that might have its own validation scheme, this API is used to convey the errors from the external validation.
+
+  The `ValidationError` contains a `type`, `path`, and `message`.
+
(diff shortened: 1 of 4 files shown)


josdejong:
Thanks for your PR @chaitu-tk .

It feels a bit like this API duplicates the custom validation API that is already available. I want to be careful introducing overlapping features. 

Can you explain your use case? I would think with the existing API you should be able to do what you describe. You can implement a custom `onValidate` function, decide whether or not this function returns anything based on a certain state of your form. You can trigger the `editor.validate()` method any moment you want (after say the user clicked Save or so), to force validation again.

chaitu-tk:
Thanks, Jose. I did exactly what you are suggesting (skip custom validation until save is clicked and return errors (if any) on save) and it works fine. I have raised this PR before I could get the custom validation to do what I want, so, feel free to close this. I don't see any use case that doesn't do custom validation and relies on external validation.

josdejong:
ah, good to hear 👍 

I'll close this PR then. Thanks for your inputs!

PR #1385, opened 2021-10-30
fix: verify object prototype before define 'remove' property in it

caok2709:
Before define 'remove' property in the item's prototype, we need to check whether item is undefined or not.
If we don't check item, it could lead to an error like:
``Cannot use 'in' operator to search for 'remove' in undefined ``
For example, window.DocumentType.prototype can be undefined in some rare environment.

### DIFF
--- a/src/js/polyfills.js
+++ b/src/js/polyfills.js
@@ -3,17 +3,19 @@ if (typeof Element !== 'undefined') {
   // Polyfill for array remove
   (() => {
     function polyfill (item) {
-      if ('remove' in item) {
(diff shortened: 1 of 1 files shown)


caok2709:
@josdejong Will you be able to merge my PR to fix this tiny issue?
Thx

josdejong:
Thanks @caok2709, sounds good to make this function more robust 👍

josdejong:
@caok2709 you're fix is now published in `v9.5.7`. Thanks again.

PR #1409, opened 2022-01-09
Feature/on node name

brianpos:
This is an alternative implementation of what was requested in PR #839 
which supports the inclusion of access to the value of the node applicable to the event.
(also updated the documentation too)

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -126,7 +126,7 @@ Constructs a new JSONEditor.
   Set a callback function triggered right after the mode is changed by the user. Only applicable when
   the mode can be changed by the user (i.e. when option `modes` is set).
 
-- `{function} onNodeName({ path, type, size })`
+- `{function} onNodeName({ path, type, size, value })`
 
   Customize the name of object and array nodes. By default the names are brackets with the number of childs inside,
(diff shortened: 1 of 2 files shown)


josdejong:
Thanks Brian, looks good. 

I was rethinking my remark about `this.value` and `this.getValue()` in https://github.com/josdejong/jsoneditor/pull/839#issuecomment-950122188. Right now, `this.value` is simply not defined for objects/arrays. It is used like this by `onEditable` and `onClassName`, so the current implementation will be consistent with that. 

But we have to decide whether that is the behavior we actually want 🤔. We can address that in a separate PR. I think it would be best to have the real value for arrays/objects too, but like you already noted before, that can cause quite a performance penalty. That would be a pity especially since you do not always _need_ the value. I think what we can do is change the `value` property into [a lazy getter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get  ) for all of the callbacks. What do you think Brian?

brianpos:
The lazy getter sounds like a good idea, will keep the API consistent, and only take the hit if it's ever actually used.

josdejong:
Yes indeed. I'll merge your PR now, thanks for implementing this Brian :). After that, I will change `value` into a getter for all relevant callbacks.

Issue #1217, opened 2021-01-09
Event to get notified when a node gets expaneded


belaviyo:
Hello,

I was wondering if there is an event to get notified when a node gets expanded or collapsed. I need this to write a custom function to keep the list of expanded states. I couldn't find any function to restore expanded nodes after a page refresh

josdejong:
Thanks for your suggestion. There is no such event right now. I think if we would add it, we will some other related methods as well to make this feature complete, such as `editor.getExpandedNodes()`, `editor.setExpandedNodes()`.

Can you explain your use case? If we have a clear view on that, we can think though a good solution.

belaviyo:
the idea is to restore expanded states when browser restarts or when the page reloads

josdejong:
That makes sense indeed. Then, maybe methods like `getState()` and `setState()` makes sense. It could contain the expanded nodes, but maybe also the scroll position and the selection.

belaviyo:
That would be perfect!

josdejong:
Sounds good.

Anyone interested in implementing a `getState` and `setState`? I think it requires some thinking though of the details before actually implementing something. Help would be very welcome.

josdejong:
Thanks for your input @himanshu520 . I like your proposal of an `onExpand(path, isExpanded)` and `editor.expand(path, isExpand)`, this is much simpler to implement than a big, generic `getState` and `setState`.

Both these methods do not yet exist, the only thing that comes close is the `editor.exapndAll()` function 😄 

I'm not planning on implementing this myself any time soon, help implementing this would be welcome.

PR #1440, opened 2022-05-18
Add support to sync node expand/collapse operations across editors

himanshu520:
* Added onExpand callback option to be specified while creating JSONEditor, that is invoked when a node is expanded/collapsed.
* Added JSONEditor.expand API that allows to expand/collapse JSON node by specifying its path.

The above two additions would allow library users to sync node expand/collapse operations when having multiple editors  (#1217).


josdejong:
Thanks @himanshu520 , your PR looks good at first sight. I'll review it soon.

docs/api.md
```diff
@@ -679,6 +693,24 @@ Destroy the editor. Clean up DOM, event listeners, and web workers.
 
 Expand all fields. Only applicable for mode 'tree', 'view', and 'form'.
 
+#### `JSONEditor.expand(path, isExpand, recursive)`
```
josdejong, line 696: Thinking aloud here: it may be nice to change the signature 

```js
JSONEditor.expand(path, isExpand, recursive) 
```
to
```js
JSONEditor.expand({path, isExpand, recursive}) 
```

Then you can pass the options even more easily from one editor to an other:

```js
const editor1 = new JSONEditor(container1, { onExpand: (props) => editor2.expand(props) })
const editor2 = new JSONEditor(container2, { onExpand: (props) => editor1.expand(props) })
```

What do you think?

josdejong (commented):
The PR looks good, exactly what we're looking for 👌. Thanks for the clear description in the docs. 

I made one inline comment, and I have one more question: can you add a new, minimal example in the `/examples` section demonstrating two editors which keep their expanded state in sync? Is useful for users and also for you and me manually testing whether all works as expected.

himanshu520:
Thanks for reviewing. I have updated the PR as per suggestions, please have a look again.

PS - I would like to add that though `editor.expand()` and `options.onExpand()` helps to sync node expanded/collapsed states across editors, my main motive was to use it to show **JSON diff** using two editors where **common/modified nodes are always aligned** (shown side-by-side, as most text-based-diff tools). However, I later realised that it is not possible to handle the scenario when there are additonal/missing nodes - then the alignment of common/modified nodes cannot be maintained. I guess, I would have to write something from scratch 😐️ but hope that the `editor.expand()` and `options.onExpand()` APIs would be useful for others 😀️

josdejong:
Thanks @himanshu520 , looks good 👌 . 

You're right, expanding only works if a path exists. I'll add a remark on top in the example explaining this.

josdejong:
Published now in `v9.8.0`

Issue #1371, opened 2021-09-03
Support Ajv v8


msalahz:
Currently, Jsoneditpr only supports Ajv v6.12.6 which is behind with two major releases, so can you please support Ajv v8?
Here is the [migration document from v6 to v8](https://ajv.js.org/v6-to-v8-migration.html)

josdejong:
Yes, would be nice to upgrade to the latest version of Ajv.

Anyone interested in helping out here?

StephenAtCFA:
Started looking into this in order to be able to support the discriminator property (in ajv v8). Ended up getting this error and I'm not familiar enough with this project to figure it out quickly. ```
main.jss:4287 Uncaught TypeError: Cannot read properties of undefined (reading '0')
    at _parsePath (main.jss:4287:500)
    at Node.findNode (main.jss:2806:60)
    at findNode (main.jss:3878:141)
    at Array.map (<anonymous>)
    at JSONEditor.treemode.validate (main.jss:3878:103)
    at JSONEditor.treemode._validateAndCatch (main.jss:3880:474)
    at JSONEditor.treemode.set (main.jss:3763:1)
    at JSONEditor.setMode (main.jss:1675:104)
    at JSONEditor._create (main.jss:1648:213)
    at new JSONEditor (main.jss:1627:209)
```

josdejong:
Thanks @StephenAtCFA . Looking at the stacktrace, something goes wrong in the `validate` function. If I remember well, the new version of Ajv has a different data model for the validation errors (like `error.dataPath` doesn't exist anymore). I think the function `improveSchemaError` needs to be adjusted for the new model.

Here is the piece of code that is triggered looking at your stacktrace:

https://github.com/josdejong/jsoneditor/blob/f5b30462d9034dcd058c2241b6c1d2e4af8fce68/src/js/treemode.js#L567-L584

howardjones:
I've run into an ajv bug that seems to be fixed in v8, so I also have been looking at updating jsoneditor. Looks like .dataPath is now .instancePath and there is no longer a .schema property? (output below from the ajv-cli for  ajv 8.11.0)

```[
  {
    instancePath: '/dns_zone_name',
    schemaPath: '#/properties/dns_zone_name/pattern',
    keyword: 'pattern',
    params: { pattern: '^[a-z0-9\\.\\-]+$' },
    message: 'must match pattern "^[a-z0-9\\.\\-]+$"'
  }
]

[
  {
[...]

PR #1472, opened 2022-10-02
Ignore unknown format when schema compilation and always pass validation

JonnyLeiZhang:
Hi @josdejong,
Currentlty more and more customized format are used in JSON schema, some of them are not  supported by AJV 6.12.6. 
When we're using these schemas, it will report error like:
Unknown format 'unit16' is used in schema validation in console.
And then the JSONEditor cannot work anymore.
I do notice there're several similar cases, I guess the straight solution is to remove these formats in the schema file. Can we directly add option {unknownFormats: 'ignore'} when we instantiate the AJV object?

Or we can make it configurable?

REF: https://github.com/ajv-validator/ajv/tree/v6.12.6#options

BR/Lei

### DIFF
--- a/src/js/JSONEditor.js
+++ b/src/js/JSONEditor.js
@@ -357,7 +357,8 @@ JSONEditor.prototype.setSchema = function (schema, schemaRefs) {
           allErrors: true,
           verbose: true,
           schemaId: 'auto',
-          $data: true
+          $data: true,
+          unknownFormats:'ignore'
         })
 
         // support both draft-04 and draft-06 alongside the latest draft-07


josdejong:
Thanks for your input @JonnyLeiZhang .

Probably most issues will be resolved by upgrading to Ajv@8, see #1371.

I think it is not a good idea to configure Ajv to ignore unknown formats by default, it's better to know when something is wrong than silently ignoring it. Right now, you can configure Ajv yourself by configuring the option `ajv`, so if you need the option `{unknownFormats: 'ignore'}`, you can do so already. See docs on `ajv`: https://github.com/josdejong/jsoneditor/blob/develop/docs/api.md#configuration-options. 

Does that make sense?

JonnyLeiZhang:
Ah, I miss that it's already configurable. Thank you!

BR/Lei

PR #1497, opened 2023-02-15
Control default behaviour of error table through options

magedhennawy:
Solves [Display error table through options if mode is other than text](https://github.com/josdejong/jsoneditor/issues/1496)

### DIFF
--- a/src/js/textmode.js
+++ b/src/js/textmode.js
@@ -50,7 +50,7 @@ textmode.create = function (container, options = {}) {
   options.enableTransform = options.enableTransform !== false
   options.createQuery = options.createQuery || createQuery
   options.executeQuery = options.executeQuery || executeQuery
-
+  options.errorTableVisible = options.errorTableVisible
   this.options = options
 
   // indentation
@@ -327,7 +327,7 @@ textmode.create = function (container, options = {}) {
   this._updateHistoryButtons()
 
   this.errorTable = new ErrorTable({
-    errorTableVisible: this.mode === 'text',
+    errorTableVisible: this.options.errorTableVisible,
     onToggleVisibility: function () {
       me._validateAndCatch()
     },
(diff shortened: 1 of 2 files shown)


josdejong:
Thanks Maged for your PR. 

Right now the built-in behavior is: show the error table initially in modes `text` and `preview`, and not in `code` mode since that has icons in the gutter. 

It is a good idea to make this behavior customizable. Two thoughts:

1. This option should allow for the current behavior. I think we could make this option a callback like
    `showErrorTable: (mode: string) => boolean` to achieve that, or a list with modes like `showErrorTable: ['text', 'preview']` (maybe also allowing a boolean). 
3. I think the option should have a good default, I think the current behavior would be good.

What do you think?

magedhennawy:
@josdejong Hey Jos, just updated it to match the showErrorTable array, and also possible a boolean. 
Please let me know what you think~

src/js/textmode.js
```diff
@@ -50,7 +50,7 @@ textmode.create = function (container, options = {}) {
   options.enableTransform = options.enableTransform !== false
   options.createQuery = options.createQuery || createQuery
   options.executeQuery = options.executeQuery || executeQuery
-
+  options.showErrorTable = options.showErrorTable || ['text']
```
josdejong, line 53: Can you change the default to `['text', 'preview']`?

josdejong (commented):
That looks good👌. I made one small inline comment.

Can you also describe the new feature in `/docs/api.md`?

magedhennawy:
All done @josdejong

josdejong:
Thanks Maged! Merging your PR now. I see there is a minor linting issue, I'll fix that directly.

josdejong:
Published now in `v9.10.0`, thanks again!

Issue #1208, opened 2020-12-29
Could not load source file "0" in source map of "../node_modules/jsoneditor/dist/jsoneditor.min.js"


hoogw:
I got this warning when build, using "jsoneditor": "^9.1.6",

                     Could not load source file "0" in source map of "../node_modules/jsoneditor/dist/jsoneditor.min.js"

full code:

                       parcel build --no-minify --no-cache --public-url ./ arcgis/*.html

                     ⚠️  Could not load source file "0" in source map of "../node_modules/jsoneditor/dist/jsoneditor.min.js".
                     ⚠️  Could not load source file "0" in source map of "../node_modules/jsoneditor/dist/jsoneditor.min.js".
                     ✨  Built in 8.14s.

josdejong:
How can I reproduce this issue?

hoogw:
how to reproduce issue:




                              npm init

                              npm install jsoneditor


I use parcel.js to bundle.   install parcel.js by: 


                              npm install --save-dev parcel-bundler     


 create a simple html, js file.
in js file, I will import jsoneditor as: 

                                

                             // jsoneditor
                                import 'jsoneditor/dist/jsoneditor.css';   // do not forget CSS !
                                import JSONEditor from "jsoneditor";


Then, run:

                     parcel build --no-minify --no-cache --public-url ./    *.html



Then you will see warning

                          Could not load source file "0" in source map of "../node_modules/jsoneditor/dist/jsoneditor.min.js".
                 ⚠️  Could not load source file "0" in source map of "../node_modules/jsoneditor/dist/jsoneditor.min.js".
                 ✨  Built in 8.14s.

This is my package.json, you could have yours, 

                                                     {
                     "name": "ol2",                          
                     "version": "1.0.0",
                     "description": "",
                     "main": "index.js",
                            "scripts": {
                              "test": "echo \"Error: no test specified\" && exit 1",
                              "start": "parcel arcgis/*.html",
                              "build": "parcel build --no-minify --no-cache --public-url ./ arcgis/*.html"
                            },
                            "author": "",
                            "license": "ISC",
                            "dependencies": {
                              "@mdi/font": "^5.8.55",
                              "jquery": "^3.5.1",
                              "json": "^10.0.0",
                             "jsoneditor": "^9.1.6",
                             "ol": "^6.4.3",
                              "ol-layerswitcher": "^3.8.3",
                              "ol-mapbox-style": "^6.3.0",
[...]

PR #1499, opened 2023-02-16
Fix source map paths in Uglify configuration

joshkel:
Source maps were written with `"0"` as the source filename.  This could result in errors similar to the following in Webpack (depending on the Webpack configuration):

```
WARNING in ../../node_modules/jsoneditor/dist/jsoneditor.min.js
Module Warning (from ../../node_modules/source-map-loader/dist/cjs.js):
Failed to parse source map from '/Users/josh/app/node_modules/jsoneditor/dist/0' file: Error: ENOENT: no such file or directory, open '/Users/josh/app/node_modules/jsoneditor/dist/0'
```

Uglify configuration based on https://github.com/mishoo/UglifyJS#api-reference.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -102,7 +102,7 @@ const compilerMinimalist = webpack({
 
 function minify (name) {
   const code = String(fs.readFileSync(DIST + '/' + name + '.js'))
-  const result = uglify.minify(code, {
+  const result = uglify.minify({ [name + '.js']: code }, {
     sourceMap: {
       url: name + '.map'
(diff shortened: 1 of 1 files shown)


josdejong:
Thanks! So this will fix #1208?

joshkel:
@josdejong, yes, this should fix #1208.

josdejong:
Awesome 😎

josdejong:
Fix published now in `v9.10.0`, thanks again!

PR #1549, opened 2023-09-18
[(develop)] fixed parseString when it trimming zero in the beggining

DarkFisk:
Hello,
I have a problem when I add new elements to an array, and even when I provide schema validation that only strings are allowed, JsonEditor always removes 0 in front of the number, after some investigation I found that `parseString` is casting values like **0**102031234 to number and removes zero in front and making 102031234, that is absolutely wrong.  (this number is actually example of CPR number in Denmark, it is like your passport number)
Please, take into consideration my PR.

P.S. @josdejong great tool, thank you!

### DIFF
--- a/src/js/util.js
+++ b/src/js/util.js
@@ -1131,6 +1131,10 @@ export function parseString (str) {
     return false
   }
 
+  if (str.length > 1 && str.charAt(0) === '0') {
+    return str;
+  }
+
   const num = Number(str) // will nicely fail with '123ab'
   const numFloat = parseFloat(str) // will nicely fail with '  '
   if (!isNaN(num) && !isNaN(numFloat)) {
--- a/src/js/util.js
+++ b/src/js/util.js
@@ -1132,7 +1132,7 @@ export function parseString (str) {
   }
 
   if (/^0\d+$/.test(str)) { // to treat '001' as a string
-    return str;
+    return str
   }
 
   const num = Number(str) // will nicely fail with '123ab'
--- a/src/js/util.js
+++ b/src/js/util.js
@@ -1131,7 +1131,7 @@ export function parseString (str) {
     return false
   }
 
-  if (str.length > 1 && str.charAt(0) === '0') {
+  if (/^0\d+$/.test(str)) { // to treat '001' as a string
     return str;
   }
 
--- a/test/util.test.js
+++ b/test/util.test.js
@@ -315,6 +315,7 @@ describe('util', () => {
     assert.strictEqual(parseString('false'), false)
     assert.strictEqual(parseString('+1'), 1)
     assert.strictEqual(parseString('01'), '01')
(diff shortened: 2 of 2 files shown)


josdejong:
Thanks for bringing this up Viacheslav. 

I think to work out a solution for this we have to look a bit deeper. You're PR only checks for values with a single leading zero, like `0123`, but if we make a solution for this we should also take care of multiple leading zeros like `000123`.

When you open a JSON document and it contains a string like `0102031234 `, it will be loaded with type "string" (not "auto"), and leading zeros will be maintained. This is not the case though wen creating a new value and pasting the value `0102031234` there. So I think we should address that logic. You can have a look at the [`setValue` method](https://github.com/josdejong/jsoneditor/blob/228dd2edbe3e7b09be2c4e3a12b755bbbdce4169/src/js/Node.js#L427) to get an idea. What do you think?

On a side note: this issue has been addressed already in the successor of this library, [`svelte-jsoneditor`](https://github.com/josdejong/svelte-jsoneditor/).

josdejong:
Thinking about this more I think it would be proper to solve this in `parseString`, since a number with leading zeros is not valid in JSON, so it has to be a string there.

Can you add unit tests to verify that `0.3` and `0e3` are still parsed as numbers? Those also start with a zero but are valid numeric values.

DarkFisk:
Thank you [josdejong](https://github.com/josdejong) for quick response,
I added one more test to PR to show you that my fix will treat value as string type no matter how many zero is in front of value.

regarding `setValue` you mentioned, it call function `_getType`, and it inside also call buggy `parseString` method, and it returns for string `0102031234` number without zero in front, but `_getType` logic returns `string` type if value is `number`, and `auto` for `string`... I went to crazy house)

DarkFisk:
> Thinking about this more I think it would be proper to solve this in `parseString`, since a number with leading zeros is not valid in JSON, so it has to be a string there.
> 
> Can you add unit tests to verify that `0.3` and `0e3` are still parsed as numbers? Those also start with a zero but are valid numeric values.

you are right, it is not so easy if such cases also need to be supported, will take a look if it possible to handle such cases

josdejong:
Ah, this regex looks solid indeed, thanks 👌 

I see the linter complains about the semicolon `;`, can you `npm run format` to fix the linting issue? Besides that we can merge your fix.

DarkFisk:
@josdejong thank you. I fixed lint error. if it will be merged, just wandering, when are you planning to release new version?

josdejong:
I'm publish the fix right away, there aren't other pending fixes/features right now.

josdejong:
Published now in `v9.10.3`, thanks again Viacheslav!

DarkFisk:
> Published now in `v9.10.3`, thanks again Viacheslav!

thank you very much, it is awesome news

Issue #1570, opened 2024-02-08
Spanish Grammar corrections in i18n file


gonchesan:
Hello Jos!!
I foun some words are missing accent marks and one of the values is still in English.

- [ ] autoType 
- [ ] modeCodeText
- [ ] modeTreeText
- [ ] containsInvalidProperties
- [ ] containsInvalidItems

josdejong:
Thanks for reporting. Can you help provide the right Spanish texts?

gonchesan:
Yes! of course! I have no problem.

josdejong:
Fixed now in `v10.0.1` via #1571.

PR #1571, opened 2024-02-08
fix:Spanish grammar & add translation in autoType 

gonchesan:
- Fix Spanish Grammar corrections in i18n file #1570


josdejong:
Thanks for the fixes! They're published now in `v10.0.1`.

PR #1642, opened 2025-02-10
Fix test suite in Node 22

joshkel:
Node now provides a `navigator` global (https://nodejs.org/api/globals.html#navigator_1), which could cause JSONEditor's test suite to fail with the following error.

```
 Exception during run: TypeError: Cannot set property navigator of #<Object> which has only a getter
    at setUpTestEnvironment (jsoneditor/test/setup.js:16:19)
    at Object.<anonymous> (jsoneditor/test/setup.js:22:1)
```


josdejong:
Thanks Josh👍

Issue #1259, opened 2021-03-03
Tree mode paste problem


zhangjin747:
In the tree mode, after paste plain text in fields, the pasted text was wrapped by a 'span' element and it cause the text cannot be selected or insert cursor.
``` html
<div contenteditable="true" spellcheck="false" class="jsoneditor-value jsoneditor-string jsoneditor-color-value" title>
    <span>gold</span>
</div>
```

josdejong:
Thanks for reporting @zhangjin747. How can I reproduce this issue exactly? And what browser are you using?

zhangjin747:
@josdejong You just need to copy some text from key or value field and paste it to another field in tree mode. And then,  the cursor can not be inserted between letters.  I'm using the latest Microsoft Edge browser based on Chromium.

josdejong:
Hm I've tried on Edge but am not able to reproduce this. Can it be related to specific locales or keyboard inputs?

zhangjin747:
@josdejong  I have figured out the problem. What I have copied seems to be formatted text(although it looks just like plain text) and if I use _paste as plain text_, everything works fine. 
In the older version, the formatted text will be converted to plain text on blur, but in the latest version, it won't happen.  I don't know if this is a bug.

josdejong:
Ahh, that helps a lot. Looking at the history, it may be related to the fix made in `v8.6.8` in the following commit:  https://github.com/josdejong/jsoneditor/commit/ad021ec7afa65da05f347b40b783947eab82abbd. 

Without being able to reproduce this myself I'm not sure though how to create a fix for the issue you encounter. Any ideas?

pawfrolow:
Hi, @josdejong  

I still see this problem on `v10.1.2`

I made some example to reproduce:
https://codesandbox.io/p/sandbox/293922

To see the problem, you need to select all the text value of "value" field and paste to "7405242042266046865". After that if you blur from that field you can't focus on it by click.

This problem disappears if I back to version `v8.6.6`

josdejong:
Thanks Pavel for looking into this. I tried to reproduce the issue following the steps that you describe for your example but still no luck reproducing the issue. I tried in Chrome and Firefox, on Windows 11. Do you have more details on how to reproduce the issue?

pawfrolow:
[...]

PR #1646, opened 2025-03-11
Stringify for nums bigger than max safe integer

pawfrolow:
Hi, I've made fixes for two problems, discussed here https://github.com/josdejong/jsoneditor/issues/1259

1: Trying to paste big number like `7405242042266046865`, that incorectly converts it to number, because it bigger that max safe integer.
Before: https://github.com/user-attachments/assets/7869e787-c2da-4d1c-ac20-0e7243720148
After: https://github.com/user-attachments/assets/be9029f3-b726-4eb5-8c52-3e9bf4cfd798

### DIFF
--- a/src/js/Node.js
+++ b/src/js/Node.js
@@ -2235,7 +2235,7 @@ export class Node {
     if (domField) {
       if (this.fieldEditable) {
         // parent is an object
-        domField.contentEditable = this.editable.field
+        domField.contentEditable = this.editable.field ? 'plaintext-only' : false
         domField.spellcheck = false
         domField.className = 'jsoneditor-field'
       } else {
@@ -2428,7 +2428,7 @@ export class Node {
       } else {
         // create an editable or read-only div
         domValue = document.createElement('div')
-        domValue.contentEditable = this.editable.value
+        domValue.contentEditable = this.editable.value ? 'plaintext-only' : false
         domValue.spellcheck = false
         domValue.innerHTML = this._escapeHTML(this.value)
       }
--- a/src/js/Node.js
+++ b/src/js/Node.js
@@ -2235,7 +2235,7 @@ export class Node {
     if (domField) {
       if (this.fieldEditable) {
         // parent is an object
-        domField.contentEditable = this.editable.field ? 'plaintext-only' : false
+        domField.contentEditable = this.editable.field
         domField.spellcheck = false
         domField.className = 'jsoneditor-field'
       } else {
@@ -2428,7 +2428,7 @@ export class Node {
       } else {
         // create an editable or read-only div
         domValue = document.createElement('div')
-        domValue.contentEditable = this.editable.value ? 'plaintext-only' : false
+        domValue.contentEditable = this.editable.value
         domValue.spellcheck = false
         domValue.innerHTML = this._escapeHTML(this.value)
       }
--- a/src/js/util.js
+++ b/src/js/util.js
@@ -145,7 +145,7 @@ const isUrlRegex = /^https?:\/\/\S+$/
 
 export function isUrl (text) {
   return (typeof text === 'string' || text instanceof String) &&
-    isUrlRegex.test(text)
+      isUrlRegex.test(text)
 }
 
 /**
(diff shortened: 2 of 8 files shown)


src/js/util.js
```diff
@@ -1144,7 +1147,7 @@ export function parseString (str) {
 
   const num = Number(str) // will nicely fail with '123ab'
   const numFloat = parseFloat(str) // will nicely fail with '  '
-  if (!isNaN(num) && !isNaN(numFloat)) {
+  if (!isNaN(num) && !isNaN(numFloat) && num <= Number.MAX_SAFE_INTEGER && numFloat <= Number.MAX_SAFE_INTEGER) {
```
josdejong, line 1150: I think the current logic doesn't work correctly for example for negative bigints like `-7405242042266046865` or valid large values like `1e25`. I think we do not have to test both `num` and `numFloat` but only one of them.

Can you improve the logic, and also write unit tests in`util.test.js`?

josdejong (changes requested):
Thanks Pavel, I think it makes sense to parse a bigint into a string rather than a number which loses digits.

I made a few inline comments, can you have a look at those?

src/js/util.js
```diff
@@ -118,7 +118,10 @@ export function getType (object) {
     return 'undefined'
   }
   if ((object instanceof Number) || (typeof object === 'number')) {
-    return 'number'
+    if (object <= Number.MAX_SAFE_INTEGER) {
```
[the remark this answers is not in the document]
pawfrolow: Ok, thanks, now I see how it works. I cancelled my changes in `getType`

src/js/Node.js
```diff
@@ -2235,7 +2235,7 @@ export class Node {
     if (domField) {
       if (this.fieldEditable) {
         // parent is an object
-        domField.contentEditable = this.editable.field
+        domField.contentEditable = this.editable.field ? 'plaintext-only' : false
```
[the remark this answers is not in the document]
pawfrolow: I moved these changes to a [separate PR](https://github.com/josdejong/jsoneditor/pull/1650)

src/js/util.js
pawfrolow: I fixed this logic with adding checks for min safe int, and adding logic for exponential numbers by regex.
Also wrote additional unit tests for parseString

pawfrolow: Thanks, now logic really looks more simplified

src/js/util.js
```diff
@@ -1144,7 +1144,10 @@ export function parseString (str) {
 
   const num = Number(str) // will nicely fail with '123ab'
   const numFloat = parseFloat(str) // will nicely fail with '  '
-  if (!isNaN(num) && !isNaN(numFloat)) {
+  const isFiniteNumber = !isNaN(num) && !isNaN(numFloat) && isFinite(num)
+  const isInSafeRange = num <= Number.MAX_SAFE_INTEGER && num >= Number.MIN_SAFE_INTEGER
+  const hasSymbols = !/^\d+$/.test(str) // for 1e12, +1, 1.23 etc
```
josdejong, line 1149: Small thing: can you rename `hasSymbols` to `isInteger` and invert it? I think that is easier to understand 😅

josdejong (commented):
Thanks for the updates Pavel, this looks good indeed! One small remark left, can you have a look?

src/js/util.js
pawfrolow: Ok, done

josdejong: 👍 can you also remove the comment? It's not matching anymore

pawfrolow: agree, it's no longer needed. I removed it

josdejong: Thanks for the updates!

josdejong (approved):
Nice! I like how this PR turned out :)

josdejong:
Ah, sorry about that, I've fixed it via d56659a2c8eebad45bac16a450d56d06a51cced3.

I've just published `jsoneditor@10.2.0` containing your new feature 🎉

PR #1651, opened 2025-03-16
Update README.md

lingceng:
I  found this chrome plugin use this repo


josdejong:
Thanks for your PR @lingceng. There are multiple Chrome Plugins and websites that use the library or its successor [`svelte-jsoneditor`](https://github.com/josdejong/svelte-jsoneditor/). Maybe someday we can create a "Showcase" section, but for now, let's not add a link to this specific plugin.

PR #1669, opened 2025-07-25
Add support for object arrays with separate `text` and `value`

ahmed-saber:
- Add support for object arrays with separate `text` and `value` properties
- Enable search functionality within both text and value fields
- Maintain full backward compatibility with existing string arrays
- Add helper functions: getOptionText(), getOptionValue(), getSearchableContent()
- Update filtering logic to handle both string and object formats
- Create comprehensive example (28_autocomplete_text_value_objects.html)
- Bump version to 10.3.0

### DIFF
--- a/examples/28_autocomplete_text_value_objects.html
+++ b/examples/28_autocomplete_text_value_objects.html
@@ -1,0 +1,166 @@
+<!DOCTYPE HTML>
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+
+  <title>JSONEditor | Auto Complete with Text/Value Objects</title>
+
+  <link href="../dist/jsoneditor.css" rel="stylesheet" type="text/css">
+  <script src="../dist/jsoneditor.js"></script>
+
+  <style type="text/css">
+    .editor-container {
+      width: 500px;
+      height: 400px;
+      margin-bottom: 30px;
+    }
+
+    p {
+      width: 800px;
+      font-family: "DejaVu Sans", sans-serif;
+      margin-bottom: 20px;
+    }
+
+    h3 {
+      font-family: "DejaVu Sans", sans-serif;
+      color: #333;
+      margin-top: 30px;
+      margin-bottom: 10px;
+    }
+
+    .demo-section {
+      margin-bottom: 40px;
+      padding: 20px;
+      border: 1px solid #ddd;
+      border-radius: 5px;
+      background-color: #f9f9f9;
+    }
+
+    .search-info {
+      background-color: #e8f4fd;
+      padding: 10px;
+      border-left: 4px solid #2196F3;
+      margin: 10px 0;
+      font-size: 14px;
+    }
+  </style>
+
+</head>
+<body>
+<h1>JSONEditor - Enhanced Autocomplete with Text/Value Objects</h1>
+
+<p>
+  This example demonstrates the enhanced autocomplete functionality that supports both traditional string arrays 
+  and new object arrays with separate <code>text</code> and <code>value</code> properties. The autocomplete can 
+  search within both the display text and the underlying value.
+</p>
+
+<div class="demo-section">
+  <h3>1. Traditional String Array Autocomplete (Backward Compatible)</h3>
+  <p>Original functionality with simple string arrays:</p>
+  <div class="search-info">
+    <strong>Available options:</strong> 'apple', 'cranberry', 'raspberry', 'pie', 'mango', 'mandarine', 'melon', 'appleton'<br>
+    <strong>Search:</strong> Type any part of the fruit names to see suggestions.
+  </div>
+  <div id="jsoneditor1" class="editor-container"></div>
+</div>
+
+<div class="demo-section">
+  <h3>2. Enhanced Object Array Autocomplete (New Feature)</h3>
+  <p>New functionality using objects with <code>text</code> (display) and <code>value</code> (actual value) properties:</p>
+  <div class="search-info">
+    <strong>Search by Company Name:</strong> Try typing "Apple", "Microsoft", "Google", etc.<br>
+    <strong>Search by Stock Symbol:</strong> Try typing "AAPL", "MSFT", "GOOGL", etc.<br>
+    <strong>Result:</strong> The dropdown shows company names, but the selected value will be the stock symbol.
+  </div>
+  <div id="jsoneditor2" class="editor-container"></div>
+</div>
+
+<div class="demo-section">
+  <h3>3. Mixed Content Example</h3>
+  <p>Countries with multiple searchable fields (name, code, and capital):</p>
+  <div class="search-info">
+    <strong>Search by Country:</strong> "United States", "Germany", "Japan"<br>
+    <strong>Search by Code:</strong> "US", "DE", "JP"<br>
+    <strong>Search by Capital:</strong> "Washington", "Berlin", "Tokyo"<br>
+    <strong>Result:</strong> Displays country name, returns country code as value.
+  </div>
+  <div id="jsoneditor3" class="editor-container"></div>
+</div>
+
+<script>
+  // Example 1: Traditional string array (backward compatible)
+  const container1 = document.getElementById('jsoneditor1')
+  const options1 = {
+    autocomplete: {
+      getOptions: function () {
+        return ['apple', 'cranberry', 'raspberry', 'pie', 'mango', 'mandarine', 'melon', 'appleton'];
+      }
+    }
+  }
+  const json1 = {
+    'fruits': ['apple', 'mango'],
+    'favorite': 'cranberry',
+    'description': 'Traditional string array autocomplete'
+  }
+  const editor1 = new JSONEditor(container1, options1, json1)
+
+  // Example 2: Object array with text/value properties
+  const container2 = document.getElementById('jsoneditor2')
+  const options2 = {
+    autocomplete: {
+      getOptions: function () {
+        return [
+          { text: 'Apple Inc.', value: 'AAPL' },
+          { text: 'Microsoft Corporation', value: 'MSFT' },
+          { text: 'Alphabet Inc. (Google)', value: 'GOOGL' },
+          { text: 'Amazon.com Inc.', value: 'AMZN' },
+          { text: 'Tesla Inc.', value: 'TSLA' },
+          { text: 'Meta Platforms Inc. (Facebook)', value: 'META' },
+          { text: 'NVIDIA Corporation', value: 'NVDA' },
+          { text: 'Netflix Inc.', value: 'NFLX' },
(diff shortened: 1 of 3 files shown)


josdejong:
Thanks Ahmed for this PR, I'll look into it next week.

josdejong:
I expect to be able to look into it coming Thursday or Friday. 

Updating the types will require a separate PR at DefinitelyTyped I think.

src/js/autocomplete.js
```diff
@@ -189,7 +222,7 @@ export function autocomplete (config) {
 
       dropDown.style.marginLeft = '0'
       dropDown.style.marginTop = element.getBoundingClientRect().height + 'px'
-      this.options = options.map(String)
+      this.options = options;
```
josdejong, line 225: I think it is important to keep this code to ensure that when a user has options containing non-string values like `[1,2,3]`, that these numeric values cannot end up being passed as a number into a custom filter function: that would be a breaking change. 

The code needs to be updated of course to only stringify values when not an object.

examples/28_autocomplete_text_value_objects.html
```diff
@@ -0,0 +1,166 @@
+<!DOCTYPE HTML>
```
josdejong, line 1: Thanks for working out an example! 

Can you also update the documentation in docs/api.md, describing that `options` can either contain an object `{text,value}` or a string. And update the documentation of function `filter` explaining that the argument `match` can be `{text,value}` or a string too.

examples/28_autocomplete_text_value_objects.html
```diff
@@ -0,0 +1,166 @@
...
+      padding: 10px;
+      border-left: 4px solid #2196F3;
+      margin: 10px 0;
+      font-size: 14px;
+    }
+  </style>
+
+</head>
+<body>
+<h1>JSONEditor - Enhanced Autocomplete with Text/Value Objects</h1>
```
josdejong, line 50: I think examples work best when they demonstrate1 thing. Can you keep only example 2 from the three examples in this file?

I have the feeling that example 3 is supposed to demonstrate more than it does right now, it refers to entering a capital like "Washington", but that seems not to work. If this example really adds value over the 2nd example, can you create a separate example file demonstrating the feature you want to demonstrate with it?

examples/28_autocomplete_text_value_objects.html
```diff
@@ -0,0 +1,166 @@
...
+  </div>
+  <div id="jsoneditor1" class="editor-container"></div>
+</div>
+
+<div class="demo-section">
+  <h3>2. Enhanced Object Array Autocomplete (New Feature)</h3>
+  <p>New functionality using objects with <code>text</code> (display) and <code>value</code> (actual value) properties:</p>
+  <div class="search-info">
+    <strong>Search by Company Name:</strong> Try typing "Apple", "Microsoft", "Google", etc.<br>
+    <strong>Search by Stock Symbol:</strong> Try typing "AAPL", "MSFT", "GOOGL", etc.<br>
```
josdejong, line 73: I tried out this example, and I can indeed search by company name (also with mixed casing), but it looks like serach by stock symbol (the value) does not work. When I enter `AA` I don't get autocompletion.

josdejong (changes requested):
Thanks for your PR @ahmed-saber . It looks well taken care of! 

I made a couple of inline comments, can you have a look at those?

Also, just as a reminder: we'll need to update the type definitions in [definitelytyped/types/jsoneditor](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/jsoneditor) after this PR is merged and published.

src/js/autocomplete.js
```diff
@@ -1,11 +1,41 @@
...
+  const searchFields = []
+  if (option.text) searchFields.push(option.text)
+  if (option.value && option.value !== option.text) searchFields.push(option.value)
+  return searchFields.length > 0 ? searchFields : ['']
+}
+
 const defaultFilterFunction = {
   start: function (token, match, config) {
-    return match.indexOf(token) === 0
+    const searchFields = getSearchableContent(match)
```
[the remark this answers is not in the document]
ahmed-saber: The code is updated now

examples/28_autocomplete_text_value_objects.html
ahmed-saber: The code is updated now

examples/28_autocomplete_text_value_objects.html
ahmed-saber: The code is updated now

src/js/autocomplete.js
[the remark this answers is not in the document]
josdejong: Thanks!

src/js/autocomplete.js
```diff
@@ -1,11 +1,41 @@
...
+  // Keep objects as-is, but convert primitives to strings to prevent breaking changes
+  return isObject(option) ? option : String(option)
+}
+
 const defaultFilterFunction = {
   start: function (token, match, config) {
-    return match.indexOf(token) === 0
+    const normalizedToken = normalizeCase(token, config)
+    return isObject(match)
+      ? normalizeCase(match.text || '', config).indexOf(normalizedToken) === 0 || normalizeCase(match.value || '', config).indexOf(normalizedToken) === 0
```
josdejong, line 31: Thanks for adding all the ` || ''` cases, that adds more robustness. I think it would be neat to move this inside the `normalizeCase` function so we don't have to repeat it this often. What do you think?

examples/29_autocomplete_multiple_fields.html
```diff
@@ -0,0 +1,111 @@
...
+    { name: 'Italy', capital: 'Rome', code: 'IT' },
+    { name: 'Spain', capital: 'Madrid', code: 'ES' },
+    { name: 'Russia', capital: 'Moscow', code: 'RU' },
+    { name: 'South Korea', capital: 'Seoul', code: 'KR' },
+    { name: 'Mexico', capital: 'Mexico City', code: 'MX' }
+  ];
+  
+  const options = {
+    autocomplete: {
+      filter: 'contain', // Use contain filter for better search experience
```
josdejong, line 63: In this object, `filter` is declared twice. I guess the line `filter: 'contain',` should be removed?

examples/29_autocomplete_multiple_fields.html
```diff
@@ -0,0 +1,111 @@
...
+
+    /* custom bold styling for non-default JSON schema values */
+    .jsoneditor-is-not-default {
+      font-weight: bold;
+    }
+  </style>
+
+</head>
+<body>
+<h1>Auto Complete with Multiple Searchable Fields</h1>
```
josdejong, line 29: This more advanced example is a really neat one, thanks for splitting it into a separate example.

examples/29_autocomplete_multiple_fields.html
```diff
@@ -0,0 +1,111 @@
...
+          value: country.code,
+          // Store additional searchable fields for custom filtering
+          _searchableFields: [country.name, country.capital, country.code]
+        }));
+      },
+      filter: function(token, match, config) {
+        // Custom filter that searches across multiple fields
+        if (typeof match === 'string') {
+          // Handle simple string matches
+          const normalizedToken = config.caseSensitive ? token : token.toLowerCase();
```
josdejong, line 77: Can you introduce a `normalizeCase` helper function in this example too (like in the built-in filter function)?

examples/28_autocomplete_text_value_objects.html
josdejong: I tried the example again, but typing the start of a value like "GOO" or "AA" still doesn't work: autocompletion doesn't show any option in that case. Does this work on your machine?

josdejong (changes requested):
Thanks for all the updates, I think the PR is almost ready to be finished. Can you have a look at the new inline comments?

examples/28_autocomplete_text_value_objects.html
ahmed-saber: This issue was resolved within the refresh function, but not in the repaint function now it should be fixed 🤞

josdejong (changes requested):
Thanks for your latest updates Ahmed. Here a few more comments.

examples/29_autocomplete_multiple_fields.html
```diff
@@ -0,0 +1,125 @@
...
+            if (typeof match === "string") {
+              // Handle simple string matches
+              const normalizedMatch = normalizeCase(match, config);
+              return normalizedMatch.indexOf(normalizedToken) > -1;
+            }
+
+            if (match && match._searchableFields) {
+              // Search across all searchable fields
+              return match._searchableFields.some((field) => {
+                const normalizedField = config.caseSensitive
```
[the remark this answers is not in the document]
ahmed-saber: Updated

src/js/autocomplete.js
```diff
@@ -104,10 +104,42 @@ export function autocomplete (config) {
           divRow.textContent = ''
           
           const rowText = getOptionText(row)
-          divRow.appendChild(document.createTextNode(rowText.substring(0, token.length)))
-          const b = document.createElement('b')
-          b.appendChild(document.createTextNode(rowText.substring(token.length)))
-          divRow.appendChild(b)
+          const rowValue = getOptionValue(row)
```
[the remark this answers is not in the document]
josdejong: Really nice, thanks 😎 . If we want we can unit test functions like `getHighlightedTextParts` (I'm fine with not adding unit tests right now though).

src/js/autocomplete.js
```diff
@@ -104,10 +104,42 @@ export function autocomplete (config) {
...
+          let matchIndex = -1
+          let matchLength = token.length
+          let displayText = rowText
+          
+          // Prefer text matches over value matches for display
+          if (rowTextLower.indexOf(tokenLower) > -1) {
+            matchIndex = rowTextLower.indexOf(tokenLower)
+            displayText = rowText
+          } else if (rowValueLower.indexOf(tokenLower) > -1) {
+            matchIndex = rowValueLower.indexOf(tokenLower)
```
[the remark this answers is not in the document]
josdejong: Ahh, now I see the dropdown shows the actual value when that is being matched and the text otherwise. That is really neat 👌. So indeed the hightlighting is needed.

ahmed-saber:
I found the line was causing `if (rows.length < 2) return` The autocomplete with `filter: 'contain'` should now work exactly as expected.

josdejong:
Published now in `v10.3.0` 🎉

ahmed-saber:
Please remember to update the version of the types as well

josdejong:
> Please remember to update the version of the types as well

I have no control over https://github.com/DefinitelyTyped/DefinitelyTyped/pull/73377, we have to await until it is reviewed and merged. I guess you'll have to use `@ts-ignore` for the time being.

PR #1674, opened 2025-08-19
Fix autocomplete text highlighting and dropdown logic

ahmed-saber:
The changes in this merge request are:
1. We always show only the label, even if the text matches one of the values
2. The dropdown will remain visible as long as the text does not match the intended value


josdejong:
Published now in `v10.4.0`.

PR #1676, opened 2025-08-22
Show dropdown for single element when option.text is available

ahmed-saber:
if you have an autocomplete option like:

`{ value: "api", text: "API Endpoint Configuration" }`

Previously: When typing "api", the dropdown would be hidden since there's only one match.
Now: The dropdown will show "API Endpoint Configuration" even though there's only one match, giving users context about what they're selecting.

image

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -9,7 +9,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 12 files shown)


josdejong:
Thanks for this fix!

josdejong:
Your fix is published now in `v10.4.1`. Thanks again 😃

PR #1758, opened 2026-06-29
fix(i18n): complete Japanese (ja) translation

mahirhir:
The Japanese (`ja`) locale in `src/js/i18n.js` is missing seven keys that exist in the English source. `translate()` falls back to English when a key is absent (`_defs[lang][key] || _defs[_defaultLang][key]`), so these labels currently render in English while the rest of the UI is Japanese:

- `compactTitle`, `formatTitle`, `repairTitle` (the Compact / Format / Repair toolbar buttons)
- `searchTitle`, `searchNextResultTitle`, `searchPreviousResultTitle` (the search box and its next/previous controls)
- `modeEditorTitle` (the editor mode switch)

Korean (`ko`) already translates all seven and Chinese (`zh-CN`) translates the first three, so this brings `ja` in line with the other locales. I kept the wording terse to match the existing Japanese entries and reused the JSON/JMESPath terms already used in the file.

Only `ja` is touched, no other keys change.

### DIFF
--- a/src/js/i18n.js
+++ b/src/js/i18n.js
@@ -476,6 +476,7 @@ const _defs = {
     ascendingTitle: '${type}の子要素を昇順に並べ替え',
     actionsMenu: 'クリックしてアクションメニューを開く (Ctrl+M)',
     collapseAll: 'すべてを折りたたむ',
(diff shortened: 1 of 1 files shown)


josdejong:
@mahirhir thanks for your PR improving the Japanese translation. 

I see you closed the PR, why is that?

PR #1789, opened 2026-09-13
fix: ignore delayed field changes on removed nodes

### LINKED ISSUE #1498: "Cannot read properties of null (reading 'getInternalPath')"
I'm sporadically getting the following error when I change object keys (not values):

```
TypeError: Cannot read properties of null (reading 'getInternalPath')
S.value
https://76uwry.csb.app/node_modules/jsoneditor/dist/jsoneditor.min.js:33:36942
eval
https://76uwry.csb.app/node_modules/jsoneditor/dist/jsoneditor.min.js:35:2194
```

See https://codesandbox.io/s/jsoneditor-debounced-error-76uwry for an example: Try changing `hello`. (This calls `setValue` extremely fast to demonstrate the problem.)

I believe that the problem is caused by a race condition: Edits are propagated with a 150ms delay (due to the use of `debounce`), so, if the editor receives a new value, the old value has been removed and no longer has the fields that the event handler expects it to have.

https://github.com/josdejong/jsoneditor/blob/v9.9.2/src/js/Node.js#L1649

I'm not sure if the correct fix is to cancel debounced events when a node is removed (e.g., Lodash's debounce has a `cancel` function; a similar idea could be added here) or make the events check for, e.g., `this.parent == null`.

sridharkalaibala:
Renaming an object key and immediately calling `editor.update()` with replacement data can remove that node before its debounced field-change callback runs. The callback then dereferences `this.parent.getInternalPath()` with a null parent and throws.

Ignore the delayed field change when the node no longer has a parent, following the approach suggested in #1498. A regression test schedules the real debounce and removes the node before it fires; a second test confirms attached-node edits still record their action and previous field value.

Validation: `npm run build-and-test` passes (build, 71 tests, StandardJS lint). Actual Chromium reproduction dispatches a key-edit input event and immediately calls `editor.update({ replacement: 42 })`: original bundle throws the reported null-parent error; rebuilt bundle produces no page error and preserves the replacement data. Diff check passes. Prepared with AI assistance and locally verified.

Fixes #1498.

### DIFF
--- a/src/js/Node.js
+++ b/src/js/Node.js
@@ -1629,6 +1629,11 @@ export class Node {
    * @private
    */
   _onChangeField () {
+    // A data update can remove this node while its field change is debounced.
+    if (!this.parent) {
+      return
+    }
+
     // get current selection, then override the range such that we can select
     // the added/removed text on undo/redo
     const oldSelection = this.editor.getDomSelection()
--- a/test/Node.test.js
+++ b/test/Node.test.js
@@ -3,6 +3,47 @@ import './setup'
 import { Node } from '../src/js/Node'
 
 describe('Node', () => {
+  describe('_onChangeField', () => {
+    it('should ignore a debounced field change after the node is removed', async () => {
+      const actions = []
+      const editor = {
+        options: {},
+        getDomSelection: () => ({}),
+        _onAction: (...args) => actions.push(args)
+      }
+      const parent = new Node(editor, { value: { hello: 'world' } })
+      const child = parent.childs[0]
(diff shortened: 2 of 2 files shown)


josdejong:
Thanks Shridhar for the PR's and issues that you've opened. I'm not sure when I can look into them but will keep you posted.

sridharkalaibala:
Thanks for the update, Jos. I appreciate you keeping me informed. I’ll leave the existing PRs ready for when you have review time and follow up on any questions or requested changes.

